PR #123, opened 2014-03-08
If boolean is stored as an integer, fix type translation for boolean

mlandauer:
Without this small fix if you try to read in a boolean that is stored
as "0" or "1" from the database you will get an error like:
undefined method `strip' for 1:Fixnum


tenderlove:
Version 2.0 removes the translation layer, so I'm going to close this. 🙇

PR #136, opened 2014-07-30
Updated the methods for pragma calls according to the currently

### LINKED ISSUE #120: Missing pragmas
There are several pragmas which do not have wrappers in `pragma.rb`.

http://sqlite.org/pragma.html

The one I noticed was `journal_mode`.

aibor:
Some of the pragmas called in the Pragmas module were deprecated and new ones were missing.
So I updated the methods for pragma calls according to the currently supported pragmas listed on https://sqlite.org/pragma.html
This should fix #120

### DIFF
--- a/lib/sqlite3/pragmas.rb
+++ b/lib/sqlite3/pragmas.rb
@@ -13,7 +13,6 @@ module SQLite3
     def get_boolean_pragma( name )
       get_first_value( "PRAGMA #{name}" ) != "0"
     end
-    private :get_boolean_pragma
 
     # Sets the given pragma to the given boolean value. The value itself
     # may be +true+ or +false+, or any other commonly used string or
@@ -39,7 +38,6 @@ module SQLite3
 
       execute( "PRAGMA #{name}=#{mode}" )
     end
-    private :set_boolean_pragma
 
     # Requests the given pragma (and parameters), and if the block is given,
     # each row of the result set will be yielded to it. Otherwise, the results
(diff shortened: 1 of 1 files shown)


tenderlove:
Honestly I'm not a huge fan of supporting all of these since they change from version to version.  It means we have to keep updating them as upstream adds more.  Could we consider just providing methods to get and set pragmas, and people can pass in the pragma name they care about?  Then we don't have to keep updating this stuff and I think it would be more clear to the user, and they can manipulate pragmas without waiting for us to upgrade our API.

aibor:
I absolutely agree that keeping the pragma methods up to date is tedious. As the different pragma type getter and setter methods are already there, it might be desirable to make them public. So if one of the wrappers is outdated because the usage has changed or new pragmas have been added, anyone can use the generic getters and setters for these pragmas.
But I think there is no point in throwing the already present wrappers over board. At least for enum pragmas I consider them rather valuable.

tenderlove:
> As the different pragma type getter and setter methods are already there, it might be desirable to make them public. So if one of the wrappers is outdated because the usage has changed or new pragmas have been added, anyone can use the generic getters and setters for these pragmas.

I agree.

> But I think there is no point in throwing the already present wrappers over board. At least for enum pragmas I consider them rather valuable.

I agree with this as well.  Can you change the PR to just make those getter / setter methods public?

aibor:
Type based getters/setters have been made public with ae6c84a.

I have also added methods for recently added pragmas. In order to keep the keep it tidy I have squashed the commits regarding these wrapper methods into ee010db, which is based on current master, now.

I'm sorry for the mess, but I hope this will keep the commit log sane.

tenderlove:
@aibor great, thank you!

PR #142, opened 2014-10-23
Added a FAQ in Markdown

penguinpowernz:
Not sure if you actually want this (another thing to maintain?) but it helped me out.  Pull it if you want it.


penguinpowernz:
Looks like the build for 1.8.7 failed because sqlite-dev was not installed.

PR #153, opened 2015-04-11
Add Database#threadsafe? to return true when SQLITE_THREADSAFE is 1 or 2

rafaelsales:
My application depends on SQLite being threadsafe. I think checking for compiler flags is too low-level, so I thought it would be useful to have this on this gem.


tenderlove:
I think [this function](https://github.com/sparklemotion/sqlite3-ruby/blob/master/lib/sqlite3.rb#L14) covers the use case.  Let me know if not and I'll reopen this.  Thank you

Issue #155, opened 2015-04-22
return symbolized keys when using results_as_hash = true


audibleblink:
Currently returning thousands of records with string instances as keys. Seems wasteful.

``` ruby
[12] pry(main)> DB.execute("select * from politicians")[0].keys.first.object_id # is String: "name"
=> 70139106472200
[13] pry(main)> DB.execute("select * from politicians")[1].keys.first.object_id # is String: "name"
=> 70139116927900
```

Issue #154, opened 2015-04-22
superfluous work in #ordered_map_for


audibleblink:
exploding zipped rows to flatten them is unnecessary since `Hash[]` takes a 2-d array

``` ruby
def ordered_map_for columns, row
  h = Hash[*columns.zip(row).flatten]
  row.each_with_index { |r, i| h[i] = r }
  h
end

def ordered_map_for2 columns, row
  h = Hash[columns.zip(row)]
  # columns.zip(row) {|k, v| h[k.to_sym] = v}
  row.each_with_index {|r, i| h[i]=r}
  h
end


require 'benchmark'

columns = %w(a b c d e f g h i j k l m n o p q r s)
row     = %w(1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9)

n = 100000
Benchmark.bm do |x|
  x.report("old ") { n.times do ; ordered_map_for columns, row;  end }
  x.report("new ") { n.times do ; ordered_map_for2 columns, row; end }
end
#######################
       user     system      total        real
old   2.730000   0.040000   2.770000 (  2.826577)
new   2.200000   0.030000   2.230000 (  2.062929)
[Finished in 5.3s]
```

PR #156, opened 2015-04-22
return symbolized keys when using results_as_hash

audibleblink:
- remove superfluous work when zipping  column
  headers and row data

Addresses #154 and #155 

Also, what purpose does inserting numbered keys serve? Just wondering because commenting it out in benchmarks makes a huge difference.

``` ruby

def ordered_map_for columns, row
  h = Hash[*columns.zip(row).flatten]
  row.each_with_index { |r, i| h[i] = r }
  h
end

def ordered_map_for2 columns, row
  Hash[columns.zip(row)]
end


require 'benchmark'

columns = %w(a b c d e f g h i j k l m n o p q r s)
row     = %w(1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9)

n = 500000
Benchmark.bm do |x|
  x.report("old ") { n.times do ; ordered_map_for columns, row;  end }
  x.report("new ") { n.times do ; ordered_map_for2 columns, row; end }
end

###########
       user     system      total        real
old  14.130000   0.240000  14.370000 ( 14.622877)
new   6.270000   0.060000   6.330000 (  6.445046)
[Finished in 21.2s]
```

### DIFF
--- a/lib/sqlite3/database.rb
+++ b/lib/sqlite3/database.rb
@@ -582,8 +582,9 @@ Support for this will be removed in version 2.0.0.
     private
 
     def ordered_map_for columns, row
-      h = Hash[*columns.zip(row).flatten]
-      row.each_with_index { |r, i| h[i] = r }
+      h = Hash.new
+      columns.zip(row) {|k, v| h[k.to_sym] = v}
+      row.each_with_index {|r, i| h[i] = r}
       h
     end
   end


tenderlove:
> Also, what purpose does inserting numbered keys serve?

The numbers are vestigial from when Ruby's hashes were not ordered.  I'm fine with this, but it does change the API in a backwards incompatible way, so we'll have to wait for a major rev.  Good news is that if we do a major rev we can drop 1.8 support and eliminate the numbered keys.

PR #165, opened 2015-09-30
Accept Time objects as bind params

alexdowad:
SQLite has 'datetime' columns, so it seems natural to pass Time objects in as bind
params when executing a query.

It would also be good to support Date and DateTime objects (from 'datetime' in
stdlib) in the future. These are not built-in Ruby core classes, so there is no
global variable like 'rb_cTime' which can be used to refer to them.

### DIFF
--- a/ext/sqlite3/statement.c
+++ b/ext/sqlite3/statement.c
@@ -282,8 +282,14 @@ static VALUE bind_param(VALUE self, VALUE key, VALUE value)
       status = sqlite3_bind_null(ctx->st, index);
       break;
     default:
-      rb_raise(rb_eRuntimeError, "can't prepare %s",
-          rb_class2name(CLASS_OF(value)));
+      if (CLASS_OF(value) == rb_cTime) {
+        /* convert to String and make a recursive call */
+        VALUE timestr = rb_funcall(value, rb_intern("strftime"), 1,
+          rb_str_new2("%Y-%m-%d %H:%M:%S"));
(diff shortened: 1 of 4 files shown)


alexdowad:
Added a couple (unrelated) commits to "clean up" warnings which occur when running the test suite... please let me know if you would like those as a separate PR instead.

alexdowad:
The build is failing on MRI 1.8.7 only, same as `master`. Strange; I can build and test successfully on 1.8.7, but Travis can't. I'd debug the problem, if only it could be duplicated locally.

tenderlove:
[The documentation says](https://www.sqlite.org/datatype3.html):

> SQLite does not have a storage class set aside for storing dates and/or times.

The documentation goes on to say:

> the built-in Date And Time Functions of SQLite are capable of storing dates and times as TEXT, REAL, or INTEGER values

It looks like your patch makes the assumption that the dates should be stored as a string, when apparently the functions will support more types than that.  In general, I'm pretty negative on adding something to this library that libsqlite doesn't directly support (like `sqlite3_bind_null` etc).  I'd rather the user decide how they want to serialize to sqlite3's built-in types than have an opinion about it (IOW, this gem is *supposed* to be just a wrapper around the sqlite API though we have other stuff in here because of legacy).

alexdowad:
@tenderlove Thanks for the lucid explanation. I have learned something here. Closing.

PR #169, opened 2015-11-05
BLOB function result when ASCII-8BIT encoding

jimflood:
Hello,
I was unable to return BLOB values from created functions using the gem. The change in this pull request allows returning of a BLOB value from a created function.

If the encoding on the string value is ASCII-8BIT, then the return value is a BLOB instead of TEXT. This is important if binary data contains null values, or is otherwise not valid UTF-8.

### DIFF
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -322,12 +322,21 @@ static void set_sqlite3_func_result(sqlite3_context * ctx, VALUE result)
       sqlite3_result_double(ctx, NUM2DBL(result));
       break;
     case T_STRING:
-      sqlite3_result_text(
-          ctx,
-          (const char *)StringValuePtr(result),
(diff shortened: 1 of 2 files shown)


jimflood:
Here is a test I've added to show the problem:

    def test_function_return_type_round_trip
      [10, 2.2, nil, "foo", Blob.new("foo\0bar")].each do |thing|
        @db.define_function("hello") { |thing| thing }
        assert_equal [thing], @db.execute("select hello(hello(?))", [thing]).first
      end
    end

Before the fix:
  1) Failure:
SQLite3::TestDatabase#test_function_return_type_round_trip [/home/jim/workspace/sqlite3-ruby/test/test_database.rb:273]:
Expected: ["foo\u0000bar"]
  Actual: ["foo"]

PR #172, opened 2015-11-06
Keep arg values on stack to prevent segfault in user-created function

jimflood:
I do not know a lot about Ruby garbage collection internals and the use of RB_GC_GUARD, but certainly from what I've found, keeping VALUE on the stack prevents premature collection.

I can easily cause a segfault by creating a user-defined function that itself creates Ruby objects while doing its work. This fix seems to alleviate the segfaults, and I think it's because this directly puts the VALUE pointers on the stack, where garbage collection cannot miss them.

### DIFF
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -359,22 +359,17 @@ static void set_sqlite3_func_result(sqlite3_context * ctx, VALUE result)
 static void rb_sqlite3_func(sqlite3_context * ctx, int argc, sqlite3_value **argv)
 {
   VALUE callable = (VALUE)sqlite3_user_data(ctx);
-  VALUE * params = NULL;
+  VALUE params[argc]; /* RB_GC_GUARD unreliable -- keep values safely on the stack */
   VALUE result;
   int i;
 
   if (argc > 0) {
-    params = xcalloc((size_t)argc, sizeof(VALUE *));
-
     for(i = 0; i < argc; i++) {
-      VALUE param = sqlite3val2rb(argv[i]);
-      RB_GC_GUARD(param);
-      params[i] = param;
+      params[i] = sqlite3val2rb(argv[i]);
     }
   }
 
   result = rb_funcall2(callable, rb_intern("call"), argc, params);
-  xfree(params);
 
   set_sqlite3_func_result(ctx, result);
 }
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -359,17 +359,22 @@ static void set_sqlite3_func_result(sqlite3_context * ctx, VALUE result)
 static void rb_sqlite3_func(sqlite3_context * ctx, int argc, sqlite3_value **argv)
 {
   VALUE callable = (VALUE)sqlite3_user_data(ctx);
-  VALUE params[argc]; /* RB_GC_GUARD unreliable -- keep values safely on the stack */
+  VALUE * params = NULL;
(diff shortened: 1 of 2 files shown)


tenderlove:
`RB_GC_GUARD` *should* work.  Can you provide a repro that segvs?  If it's not working maybe we need to fix something upstream.

larskanis:
`RB_GC_GUARD` makes sure, that the `param` is stored on the stack and not GC'ed, but only for the scope of the variable. Since the scope is limited to the inner loop, only the current value is protected, not the whole array.

Variable length arrays as in the PR on the other hand are a C99 feature, which is not supported on MS-Visual C. They are also limited by the stack size of the process (typically 8MB on Linux).

Another (but slightly slower) alternative could be calling `rb_gc_register_address()` / `rb_gc_unregister_address()` on each array value or the usage of `rb_ary_new()` and `RARRAY_PTR` instead of `xcalloc()` .

jimflood:
I will put together a simple repro that segfaults. It will take a bit of time to find just the right function that fools the GC into releasing the relevant objects, but I should be able to find a simple case that does it.

tenderlove:
> RB_GC_GUARD makes sure, that the param is stored on the stack and not GC'ed, but only for the scope of the variable. Since the scope is limited to the inner loop, only the current value is protected, not the whole array.

Argh yes.  Totally makes sense.

@jimflood I think converting this to `rb_ary_new` rather than using `xcalloc` would be fine.

jimflood:
OK. I've been busy today, but let me find time to create a reproducible segfault, and then fix it with rb_ary_new and then update the pull request.

tenderlove:
@jimflood I don't think we need a segv repro, just the fix.  @larskanis was kind enough to point out what I was overlooking in the code.  This is definitely a bug in sqlite3-ruby.  Thank you for fixing this.

jimflood:
Well, this definitely reproduces the crash, reliably:

```
require 'sqlite3'
require 'openssl'

db = SQLite3::Database.new("test.db");

db.create_function("bug", -1) do |func, *values|
  hmac = OpenSSL::HMAC.new("1234567890123456", OpenSSL::Digest.new('sha256'))
  values.each { |value| hmac.update(value) }
  func.result = hmac.digest
end

n = Array.new(127, File.read("setup.rb"))
markers = Array.new(n.length, "?").join(",")

m = Mutex.new
threads = []
for i in 0..100
  threads << Thread.new do
    m.synchronize { puts Thread.current }
    m.synchronize { result = db.execute("select bug(#{markers})", *n).inspect }
  end
end

threads.each { |thr| thr.join }

db.close

```

$ git checkout rb_gc_guard_unreliable
$ rake install_gem
$ ruby bug.rb

and it works fine, **every** time.

$ git checkout master
$ rake install_gem
$ ruby

and it crashes, **every** time:

[crash_output.txt](https://github.com/sparklemotion/sqlite3-ruby/files/122626/crash_output.txt)

jimflood:
Comment out this line:

`      skip("Segfault is unsettling to see when running rake test")
`
in `test/test_database.rb`, and this line:

`      rb_ary_push(guard, param);
`
in `ext/sqlite3/database.c`

and you should see the segfault that this fixes. It's probably system dependent but 30 threads and 127 params of 28,000 chars each per function call hits it every time when I run it here.

I'm assuming that Ruby GC will (properly) take care of the allocated `guard` array for me.

tenderlove:
@jimflood yes, the GC should take care of the allocated array for you.

jimflood:
Good! Thx.

Issue #173, opened 2015-11-11
Bug that causes phantom disk usage


penguinpowernz:
Say I want to get the path for my database object, how can I do that?

``` ruby
db = SQLite3::Database.new("my.db")

# I want to do this
db.path   # => /home/me/my.db
db.exist? # => true
```

Then this would be good too for checking the database exists before executing a query:

``` ruby
db = SQLite3::Database.new("my.db")

FileUtils.rm "my.db"

db.execute(".tables") # => #<SQLite3::DatabaseNotFound: SQLite3::DatabaseNotFound message: "Database not found at /home/me/my.db">
```

**Because right now, if your database is deleted, this library will hold onto the filehandle, and continue adding stuff to the deleted database**.  This will cause phantom disk usage that could fill up your disk until you reboot (or delete the handle manually from `/proc`):

``` sh
$ lsof | grep '(deleted)'
ruby      21151  server    9ur     REG        8,6    24576         21 /data/summary.db (deleted)
ruby      21151  server   10u      REG        8,6   268320         18 /data/summary.db-wal (deleted)
```

This library should be more responsible than this.

If you have the same issue and want to know how to delete it then do this (recall 21151 from the above output is the PID):

``` sh
$ cd /proc/21151/fd
$ ls -lh
lrwx------ 1 server server 64 Nov 11 16:53 63 -> /data/summary.db (deleted)
lrwx------ 1 server server 64 Nov 11 16:53 64 -> /data/summary.db-wal (deleted)
$ :> 64 # truncate the wal file
$ :> 63 # truncate the database file
$ kill 21151
```

penguinpowernz:
I guess having this library checking the filesystem every time would make it quite a bit slower.

penguinpowernz:
Made a pull request for this on #174 - added a new class called SafeDatabase that checks the filesystem for DB existence.

penguinpowernz:
It should be noted that this will only happen on POSIX systems.  Others will not let an open file be deleted.

penguinpowernz:
FYI there is some discussion going on around this over in #174.

penguinpowernz:
FYI new pull request to handle this using bitmasks over on #180.

flavorjones:
#180 was merged, closing this.

PR #174, opened 2015-11-12
SafeDatabase class that checks the DB existence on the filesystem

penguinpowernz:
It has the same API as `SQLite3::Database` but `#new` and `#open` mean different things now:

* `#new` causes the database to be created, skipping the existence check
* `#open` causes the database path to be checked, raising a `SQLite3::DatabaseNotFound` exception if it doesn't exist

Before the following methods pass to `super` the check is done:

* `.execute`
* `.execute2`
* `.execute_batch`
* `.query`

Should fix the phantom disk usage bug identified in #173

### DIFF
--- a/lib/sqlite3/errors.rb
+++ b/lib/sqlite3/errors.rb
@@ -41,4 +41,5 @@ module SQLite3
   class FormatException < Exception; end
   class RangeException < Exception; end
   class NotADatabaseException < Exception; end
+  class DatabaseNotFound < Exception; end
 end
--- a/lib/sqlite3/safe_database.rb
+++ b/lib/sqlite3/safe_database.rb
@@ -1,0 +1,46 @@
+require 'sqlite3/database'
+
+module SQLite3
+  class SafeDatabase < SQLite3::Database
+
+    def self.open(path)
+      raise SQLite3::DatabaseNotFound, "Could not find database at #{path}" unless File.exist?(path)
+      self.new(path)
+    end
+
+    def initialize(path)
+      @path = path
+      super(path)
+    end
+
+    def exist?
+      File.exist?(@path)
+    end
+
+    def exist!
+      raise SQLite3::DatabaseNotFound, "Could not find database at #{@path}" unless exist?
+      true
+    end
+
+    def execute(sql, bind_vars = [], *args, &block)
+      exist!
+      super(sql, bind_vars, *args, &block)
+    end
+
+    def execute2( sql, *bind_vars )
+      exist!
+      super(sql, *bind_vars)
+    end
+
+    def execute_batch( sql, bind_vars = [], *args )
+      exist!
+      super(sql, bind_vars, *args)
+    end
+
+    def query( sql, bind_vars = [], *args )
+      exist!
+      super(sql, bind_vars, *args)
+    end
+
+  end
+end
--- a/lib/sqlite3/safe_database.rb
+++ b/lib/sqlite3/safe_database.rb
@@ -3,6 +3,8 @@ require 'sqlite3/database'
 module SQLite3
   class SafeDatabase < SQLite3::Database
 
+    attr_reader :path
+
     def self.open(path)
       raise SQLite3::DatabaseNotFound, "Could not find database at #{path}" unless File.exist?(path)
       self.new(path)


tenderlove:
The problem you identified in #173 would also occur with just a regular file handle.  Why is SQLite3 special?

penguinpowernz:
Well, it's not special... except that we are treating a file handle like a database, so we should take care to ensure that the database is always there, and if it's not raise an error to tell the calling program that it's database disappeared to:

1. ensure data integrity (by not writing to a deleted database)
2. ensure file system integrity (but not writing to a phantom file)

I think this is the responsible thing to do, but obviously it will be slower than the vanilla implementation and some people might not care, so they shouldn't be nannied by the library they are using (especially if they would rather do the check elsewhere).  I think it would be handy for people to have a class they can use that will do these checks for them, and that's why I did this PR.

jimflood:
I don't think it is the responsibility of the gem to look for deleted files. That is for the implementation using the gem. What I think would be more useful is more access to the mode flags passed to sqlite3_open_v2. For example, today, you can only test for the existance of the file on SQLite::Database.new when you include { readonly: true }, because if you don't use :readonly in the options, the C code uses sqlite_open or sqlite_open16, both of which default to SQLITE_OPEN_CREATE.

I have to work around this in order to open a database for writing, and fail if the file does not exist. In C, I would just sqlite_open_v2 with mode SQLITE_OPEN_READWRITE (i.e. leave off SQLITE_OPEN_CREATE). I think this is the appropriate level of responsibility of the gem.

penguinpowernz:
Is it not a good idea to include this class as a kind of util class to help users of the library?  Or at least include it in an example somewhere in case users of the library don't know about this potentially dangerous pitfall in SQLite3.

It's all about making the library easier to use and reason about as right now the principle of least surprise is not in effect.  IMHO this library should abstract away the inefficiencies and pitfalls of SQLite (of which there are many, LOL) to make it easier for everyone to use.

Does that make sense?

penguinpowernz:
If you decide not to merge this PR then you should at least include `.exist?` and `.path` methods on the main database class so that users can protect themselves.

jimflood:
@penguinpowernz well, I agree and disagree. I don't think that it should check for the existence of the file on every command, IMHO. But it would be nice if it checks for the existence of the file at open, using the mode flags already available from SQLite. If you had access to the full set of mode flags of sqlite_open_v2, then you could ask the open to fail if a read/write database file did not exist.

In any case, it's the maintainers of the gem who decide, not me. ;-)

tenderlove:
I don't really want to maintain this class, but I'm not against adding code to enable someone else to maintain it.

> If you decide not to merge this PR then you should at least include .exist? and .path methods on the main database class so that users can protect themselves.

Would providing just `path` be sufficient?  (Though I'll probably call it `filename`)  The other problem is that a SQLite3 connection can actually be composed of multiple files.  See [ATTACH DATABASE](https://www.sqlite.org/lang_attach.html).  This means that you need to provide the database name that you care about (though we could just default that to "main" for the `filename` method).

> If you had access to the full set of mode flags of sqlite_open_v2, then you could ask the open to fail if a read/write database file did not exist.

I agree with this too.  I think it's an API oversight that you can't provide the mode flags.  We should make it take those flags.

jimflood:
@tenderlove I would like that mode access to sqlite_open_v2 myself, so I'll take a look at the code again in database.c and propose a solution.

tenderlove:
> I would like that mode access to sqlite_open_v2 myself, so I'll take a look at the code again in database.c and propose a solution.

Thanks.  I think we can either add a third option to `initialize` on the database, or use the options hash better.  Right now [you can kind of toggle the mode](https://github.com/sparklemotion/sqlite3-ruby/blob/master/ext/sqlite3/database.c#L82-L88), but maybe just taking a bitmask integer would be less maintenance work (in case sqlite3 adds more options, we don't need to update hash keys, etc).

penguinpowernz:
> But it would be nice if it checks for the existence of the file at open, using the mode flags already available from SQLite.

Hey yea, could have an `open!` method maybe, which raises an exception if the file doesn't exist?  Most people should be used to that convention from rails `.find!` right?

> I don't think that it should check for the existence of the file on every command, IMHO

Yea that would be slow, but it would offer protection against POSIX systems which let open files be deleted.  But then there's the portability problem....  If a library uses the `SafeDatabase` class and it's running on a mac they'll get no benefit, just a speed penalty :(

> The other problem is that a SQLite3 connection can actually be composed of multiple files

Shit, I totally forgot about that.  Could possibly make `.filename` into `.filenames` and return an array? IMHO `.exist?` should be added just as a convenience method (`db.exist?` looks nice and is idiomatic) and make it check all the files in that array.  Although that's not very useful when it comes to figuring out which file is non-existent, but that possibly could be solved with a `.exist!` method which raises an error containing the filename in the message (spitballing here).

jimflood:
I think the easiest, and least disruptive, is to add an option called 'mode' to the options hash, along with Ruby constants for the mode bits used now by SQLite3.

If "mode" is not in the options hash, then you get the behavior exactly as it is today:

`int mode = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE;`

or, if `options[:readonly]`:

`mode = SQLITE_OPEN_READONLY;`

This code would not change. Existing applications see the same behavior.


If `options[:mode]`, then this sets the mode, and again, `options[:readonly]` would *add* `SQLITE_OPEN_READONLY` to this (because otherwise you might be surprised). So, `options[:mode]` is an integer, and replaces this line:

`int mode = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE;`

Adding the Ruby constants would be a nice syntax sweetener. I will code up a pull request.

penguinpowernz:
Hmmm... that solution with the the bitmask seems ambigous and not ruby like (looks more like something you'd see in C code)... but thats probably just be me. Wouldn't opening it read only mean that you couldn't write anything into the database? Seems weird.

jimflood:
@penguinpowernz In the end, I kept it simple and Ruby-like, [in this pull request](https://github.com/sparklemotion/sqlite3-ruby/pull/180).

penguinpowernz:
Closing this due to the agreed consensus, maybe we could add this class to the wiki for POSIX users who more worried about data integrity than performance?

PR #179, opened 2016-02-09
use rb_apply so we don't need to maintain two arrays

tenderlove:


larskanis:
LGTM. This also removes the possible memory leak at `params` .

PR #180, opened 2016-02-16
add 'readwrite' flag to Database.new options for readwrite without cr…

jimflood:
…eate

The SQLite3 [doc page](https://www.sqlite.org/c3ref/open.html) for `sqlite_open`, `sqlite_open16`, and `sqlite3_open_v2` reads, "The flags parameter to sqlite3_open_v2() can take one of the _following three values_ [emphasis added], optionally combined with..."

Those three values are:

```
SQLITE_OPEN_READONLY
SQLITE_OPEN_READWRITE
SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE

```

This change preserves the prior behavior of the gem, to use `SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE` unless otherwise specified, and to use `SQLITE_OPEN_READONLY` if `{ :readonly => true }`, and adds a new option, `readwrite`, to cover the third SQLite3 value, which is open for read/write without create. I think this best fits the character of sqlite3_open_v2 and the simplicity of the gem's interface.

The `{ :readwrite => true }` lets you gracefully rescue an `SQLite3::CantOpenException` similar to what you can do with today with `{ :readonly => true }`.

### DIFF
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -47,6 +47,7 @@ static VALUE initialize(int argc, VALUE *argv, VALUE self)
   VALUE opts;
   VALUE zvfs;
 #ifdef HAVE_SQLITE3_OPEN_V2
+  VALUE flags;
   int mode = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE;
 #endif
   int status;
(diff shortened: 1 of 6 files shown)


tenderlove:
Looks like the tests aren't passing.   What made you decide again using a bitmask?  I was thinking the bitmask might be nice because we won't have to deal with new options in the future.  Also, IIRC, Ruby's File API does use bitmasks (though I can't remember which method off the top of my head)

jimflood:
Right, the File API is a good example. But as I read the SQLite3 docs fresh again, it seemed simplest in the "SQLite3 way" to offer the three major options. I do like the option of a bitmask, too.

It's basically the three modes: readonly, readwrite, and readwrite+create, and then, beyond that you are mostly dealing with specialty bits.

I would like to see, for example, { flags: SQLite::OPEN_NOMUTEX }

Then you could either SQLite::Database.new(path, readwrite: true) or, SQLite::Database.new(path, flags: SQLite::READWRITE | SQLite::OPEN_NOMUTEX).

I could add the flags to this pull request if you like.

The tests that are failing are anything to do with me. I just pull from master, and they already failed. It's these three:

test_filename
test_filename_with_attachment
test_db_filename

tenderlove:
> I could add the flags to this pull request if you like.

Yes please.

> The tests that are failing aren't anything to do with me.

Ugh. I'll fix it.  We should probably put the travis badge in the README.

penguinpowernz:
>  I was thinking the bitmask might be nice because we won't have to deal with new options in the future.

Why is this a problem?  Seems like laziness and bad open source stewardship if you are hobbling the library just so you don't have to do work later?  Even if you don't do the work later, someone will submit a PR sooner or later to support new features.  I don't get it....

tenderlove:
@penguinpowernz we can't see the future, so SQLite3 may add options that don't make sense in the hash.  However, we do know that anything added will fit in the bitmask.  If we expose the bitmask it buys us time to think about "The Right Way" to expose the feature while still providing a way for users to access that new feature.  Please watch your tone.

penguinpowernz:
Doesn't seem right to me... but it's your project....

jimflood:
FYI: The flags options is immediately useful to anybody that wants to set the threading mode on open, or, for use a shared or private cache.

jimflood:
I fixed the problem in the failing filename tests, too -- the ensure block ran before the variable was set nil.

I defined all the SQLite3 open flags even those that are only meaningful with the vfs, for good style.

It should be all good to go. If this is merged in, how long from now until it makes it into the gem that comes from rubygems? When is the next release planned?

Thanks for maintaining this gem!

jimflood:
The AppVeyor errors are not mine -- they are from the filename tests. For example:

 
  1) Failure:
SQLite3::TestDatabase#test_filename [C:/projects/sqlite3-ruby/test/test_database.rb:32]:
--- expected
+++ actual
@@ -1 +1 @@
-"C:/Users/appveyor/AppData/Local/Temp/1/thing20160217-2092-lyojeg"
+"C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\thing20160217-2092-lyojeg"

lib/sqlite3/constants.rb
```diff
@@ -46,4 +46,27 @@ module ErrorCode
...
+    TRANSIENT_DB     = 0x00000400  # VFS only
+    MAIN_JOURNAL     = 0x00000800  # VFS only
+    TEMP_JOURNAL     = 0x00001000  # VFS only
+    SUBJOURNAL       = 0x00002000  # VFS only
+    MASTER_JOURNAL   = 0x00004000  # VFS only
+    NOMUTEX          = 0x00008000  # Ok for sqlite3_open_v2()
+    FULLMUTEX        = 0x00010000  # Ok for sqlite3_open_v2()
+    SHAREDCACHE      = 0x00020000  # Ok for sqlite3_open_v2()
+    PRIVATECACHE     = 0x00040000  # Ok for sqlite3_open_v2()
+    WAL              = 0x00080000  # VFS only
```
tenderlove: Can you change these to be defined in C so that we don't have to know the bitmasks directly?

jimflood:
I've updated the pull request to define SQLite3::Constants::Open in C direct from the sqlite3.h symbols.

Issue #189, opened 2016-06-26
Incorporate some of sqlite_ext gem capabilities and/or enable cleaner extension?


stevecj:
I wrote https://rubygems.org/gems/sqlite_ext to provide a means of registering function definitions to make sure they will be created on all new instances of `SQLite3::Database`. It also provides a means of asking a `SQLite3::Database` instance what functions have been created on it.

I'm thinking that it would certainly make sense to add the functionality to ask the db instance about its created functions into the sqlite3 gem.

I'm not sure that the registering of functions to be created on each new instance should be added to the sqlite3 gem, but maybe add a better way to hook into it? I'm currently `prepend`ing a module to override `initialize` and call the original `initialize` via `super`, but it would be nice to have an explicit API for hooking in that kind of thing.

tenderlove:
I am open to suggestions, but at first glance reading this I would recommend that you provide a subclass of `SQLite3::Database` and ask that users instantiate your class. That way you have full control over what functions are registered, etc.  Please feel free to open a PR if you have some suggestions for extension.

Thanks!

PR #190, opened 2016-06-28
Track names of fns created on db instance

stevecj:
I have found that it useful to be able check whether a function with a particular name has been created on an instance of `SQLite3::Database`. In the sqlite_ext gem, I monkey-patched that into the class, but I think it makes sense to have that as a native capability of the sqlite3 gem.
This PR adds that functionality.

### DIFF
--- a/lib/sqlite3/database.rb
+++ b/lib/sqlite3/database.rb
@@ -337,11 +337,19 @@ Support for this will be removed in version 2.0.0.
       self
     end
 
+    # Returns an array of the names of all functions that have been created on
+    # the target instance using #create_function.
     def created_function_names
       @created_function_names ||= []
       @created_function_names.dup
     end
 
+    # Returns true if a function with the given name (case insensitive) has been
+    # created on the target instance using #create_function.
+    #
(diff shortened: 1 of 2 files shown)


stevecj:
Note that the Travis CI failure was caused by "An error occurred while installing rake (11.2.2), and Bundler cannot continue." for Ruby 1.8.7 and seems to be unrelated to the code in this PR.

flavorjones:
I'd like to rebase this onto origin/master to ensure CI is green, but it doesn't look like I have permissions to edit your branch.

flavorjones:
It seems like `sqlite_ext` still works with modern `sqlite3` but it doesn't look like it's actively maintained.

I've sent a small PR to that repo to update the dev environment (https://github.com/stevecj/sqlite_ext/pull/1). If it's accepted then maybe we can revisit this change, but in the meantime I'm going to close it.

flavorjones:
Ah, also see @tenderlove's comment on https://github.com/sparklemotion/sqlite3-ruby/issues/189

PR #199, opened 2016-10-28
Use 'brew --prefix' for OS X/macOS compile flags

bfontaine:
This is slightly faster and doesn't rely on the format of `brew info`.


tenderlove:
Awesome, thank you!

Issue #206, opened 2017-01-07
Use #sqlite3_exec for executing multiple statements


gazayas:
`database.rb` line 268 talks about using sqlite3's api for batch execution.
I looked at the API and thought `sqlite3_exec` would be a good function to use for this.

Thoughts?

tenderlove:
We don't wrap the function today.  I'm not opposed to adding the function per-se, but I'm not sure what the win is over the current API in Ruby.  The use case is that you have a bunch of SQL statements that are semicolon separated?

gazayas:
I see, yeah, they would just be separated by semi-colons.
Like I mentioned in the PR, the only plus is that it returns values, but yeah, no type casting... 🐧

gazayas:
Pull Request #207 was merged so I'll close this issue

PR #207, opened 2017-01-08
Add new execute_batch function

gazayas:
I added an execute_batch function called #execute_batch2 which uses the sqlite3 API (specifically, the function #sqlite3_exec) to execute multiple statements.

I think it might be difficult to bind params to multiple statements with the execute_batch2 function, but if it's possible I'd definitely like to do it (seeing that #execute_batch can) so I'm marking this PR as a work in progress.
Also, I set the 3rd and 4th arguments (callback function and 1st argument to callback) of sqlite3_exec to 0.

Please let me know what you think.


gazayas:
Got the callback function to work. It'll return an array of arrays like
```ruby
[["foo"], ["bar"]]
```
when a `SELECT` statement is called as the test shows. No type translation or binding parameters though.

gazayas:
Looking at #step in Statement.c it looks like this should return all the values appropriately as well, I'll commit that here soon

gazayas:
Hmm, although it seems the callback function uses #sqlite3_column_text to return the data (which is why the 3rd argument is char **), so I'm not sure how easy it'll be to get the original type without making it into a stmt object.

ext/sqlite3/database.c
```diff
@@ -717,6 +717,55 @@ static VALUE transaction_active_p(VALUE self)
...
+  sqlite3RubyPtr ctx;
+  int status;
+  VALUE callback_ary = rb_ary_new();
+  char *errMsg;
+  VALUE errexp;
+
+  Data_Get_Struct(self, sqlite3Ruby, ctx);
+  REQUIRE_OPEN_DB(ctx);
+
+  status = sqlite3_exec(ctx->db, StringValuePtr(sql), callback_function, &callback_ary, &errMsg);
```
tenderlove: `VALUE` is already a pointer, so you can just do `callback_ary` here and remove the casts in `callback_function`

ext/sqlite3/database.c
```diff
@@ -717,6 +717,55 @@ static VALUE transaction_active_p(VALUE self)
...
+{
+  VALUE * result = (VALUE *)callback_ary;
+  VALUE row_ary = rb_ary_new();
+  int i;
+
+  for (i = 0; i < count; i++) {
+    if (data[i] == NULL){
+      rb_ary_push(row_ary, Qnil);
+    } else {
+      VALUE str = rb_str_new_cstr(data[i]);
```
tenderlove: This means that return values will always be strings where the normal query mechanism will typecast.  For example "SELECT 1" will return the string "1" with this, but the integer 1 using other APIs.  I'm not sure that this kind of inconsistency is a good thing.

test/test_database.rb
```diff
@@ -121,6 +121,32 @@ def test_batch_last_comment_is_processed
...
+        # "names" is not a valid column
+        @db.execute_batch2 'INSERT INTO bros (names) VALUES ("bazz")'
+      end
+
+      return_value = @db.execute_batch2 <<-eosql
+        CREATE TABLE more_bros (id integer PRIMARY KEY AUTOINCREMENT, name string, age integer(3));
+        INSERT INTO more_bros (age) VALUES (30);
+        SELECT * FROM more_bros;
+        eosql
+      assert_equal return_value, [["1", nil, "30"]]
```
tenderlove: This kind of illustrates the type casting issue I commented about above.

tenderlove:
> I'm not sure how easy it'll be to get the original type without making it into a stmt object.
> 
> Thoughts?

Not sure if we can.  I'll have to dig in the sqlite3 source.  Maybe rather than returning an array of arrays, we could yield to a block; that would more closely match the sqlite3 APIs.  Then clients could figure out what to do with it (do they want an array of arrays, or whatever?)

gazayas:
Thank you for the review!

Yeah, I think the type casting issue is the biggest problem here. Inconsistencies makes things more complicated. I think the best way to solve it is like you said, yielding a block so the client could handle the return values however they want to.

I totally hear you on a real world use for the function; I think the only real benefit of this one is that it DOES have a return value as opposed to #execute_batch which only returns nil, but I guess the question is, as an execute batch function, does it really need to return anything in the first place?

To be 100% honest the reason I made the function was because I saw the comment like I mentioned in Issue #206 😰  haha. As far as real world use...

tenderlove:
> I totally hear you on a real world use for the function; I think the only real benefit of this one is that it DOES have a return value as opposed to #execute_batch which only returns nil

Ah interesting, I didn't realize this.

Why don't we change this PR such that the data and columns array are yielded to a block, then lets change `execute_batch` to use this new function.  Since `execute_batch` has no return value, then using this interface might be better.  I think as long as we document that this function doesn't do any typecasting, it would be OK to introduce it.

If I'm reading `execute_batch` correctly, it doesn't support multiple statements (which I think maybe you mentioned), so we should fix that.

gazayas:
Sorry for the delay

> Why don't we change this PR such that the data and columns array are yielded to a block, then lets change execute_batch to use this new function.

Do you mean one array where the first value is the column name?
```
["name", "foo", "bar"]
```

Or 2 separate arrays?
```
[["name"], ["foo", "bar"]]
```

If the latter, the return values info will probably end up looking like this
```
[["name"], ["foo", "bar"], ["age"], ["20", "30"]]
```

Thank you! 🐧

gazayas:
Alright, I made the function return hashes with the column names and the values.

I know this doesn't leave an option to make `results_as_hashes` false, but I think it would be good to always return this as hashes so the user knows that they have to handle the values returned correctly.

gazayas:
I went ahead and created another callback function that returns the values accordingly if `results_as_hash` is false.

I would have liked to make this into one function and simply change the arity, but because the sqlite3 C library calls it directly I can't add/remove any arguments.

Let me know what you think.

Thank you

tenderlove:
Can you remove all the `.DS_Store` files please?  Also, would you mind changing the table names in the tests?  I understand the irony, but I'm afraid people reading the code might not.  Other than that, looks great!  Thank you for the work on this, I appreciate it!

gazayas:
Of course, sorry about that haha

tenderlove:
Thank you kindly!

gazayas:
No problem! Thank you too! 😄

PR #214, opened 2017-04-04
Add ruby24 to fat binaries and add mingw dependency tag

larskanis:
Update rake-compiler and rake-compiler-dock to support ruby-2.4 fat binary gems. This drops support for ruby-1.9.3 binary gems.

In the second commit this adds msys2 library dependency tag in gem metadata.
    
RubyInstaller2 supports metadata tags for installation of dependent MSYS2/MINGW libraries. The sqlite3 source gem requires the sqlite3 library to be installed on the system, which the gem installer takes
care about, when this tag is set.
    
The feature is documented here: https://github.com/oneclick/rubyinstaller2/wiki/For-gem-developers#msys2-library-dependency

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,8 +6,9 @@ source "https://rubygems.org/"
 
 
 gem "minitest", "~>5.9", :group => [:development, :test]
-gem "rake-compiler", "~>0.9.3", :group => [:development, :test]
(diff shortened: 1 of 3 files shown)


tenderlove:
@larskanis thanks Lars!

PR #234, opened 2018-06-19
Add gem build to Appveyor - update to 3.24.0

MSP-Greg:
Testing on Appveyor does almost all the work required to build a fat-binary gem.  Adding six files, three specific to sqlite3-ruby, allows creating the fat-binary gem, and also uses it for testing.

64 bit gem is 2.0 <= ruby vers <= 2.6 (trunk).  32 bit gem is 2.0 <= ruby vers <= 2.5.

See results at my fork https://ci.appveyor.com/project/MSP-Greg/sqlite3-ruby/build/2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,13 @@ ports
 InstalledFiles
 Makefile
 Gemfile.lock
+
+# windows folders & files
+*.gem
+./Rakefile_wintest
+lib/fastfilereaderext.rb
+lib/rubyeventmachine.rb
+win_gem_test/shared/local_paths.ps1
+win_gem_test/shared/local_paths.ps1.sample
+win_gem_test/packages/
+win_gem_test/test_logs/
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,36 +1,14 @@
 ---
-version: "{build}"
-branches:
-  only:
-    - master
-    - 1-3-stable
-clone_depth: 10
 install:
-  - SET PATH=C:\Ruby%ruby_version%\bin;%PATH%
-  - ruby --version
-  - gem --version
-  - gem install bundler --quiet --no-ri --no-rdoc
-  - bundler --version
-  - bundle install
+  - set BRANCH=master
+
+  # Default path uses Ruby 1.9.3 ??
+  - set PATH=C:\Ruby25-x64\bin;C:\Program Files\7-Zip;C:\Program Files\AppVeyor\BuildAgent;C:\Program Files\Git\cmd;C:\Windows\system32
+
 build_script:
-  - rake native gem
-test_script:
-  - rake test
-artifacts:
-  - path: pkg\*.gem
+  - ps: .\win_gem_test\sqlite3-ruby.ps1 $env:gem_bits
 
 environment:
   matrix:
-    - ruby_version: "193"
-    - ruby_version: "200"
-    - ruby_version: "200-x64"
-    - ruby_version: "21"
-    - ruby_version: "21-x64"
-    - ruby_version: "22"
-    - ruby_version: "22-x64"
-    - ruby_version: "23"
-    - ruby_version: "23-x64"
-    - ruby_version: "24"
-    - ruby_version: "24-x64"
-    - ruby_version: "25"
-    - ruby_version: "25-x64"
+    - gem_bits: 64
+    - gem_bits: 32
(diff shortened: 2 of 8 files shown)


tenderlove:
Do we actually need to do this?  I thought @larskanis was recommending we don't build binary gems for windows anymore.

MSP-Greg:
@tenderlove

Sorry for the second message, but I just recalled one situation that left me with 'I never thought of that'...

The issue involved what was described as an 'app', and I didn't inquire.  The person seemed to be a MacOS/*nix type, but he tested his app on Windows also.  I got the distinct impression that all he used was fat-binary gems for dependencies.

He discovered an issue with one of the dependencies.  I had built a fat-binary several commits later.  He downloaded it, installed it, and his phrase was 'it just worked'.  So, another case for someone who might not be able to build the gem, and also a case for having a fat-binary built with every commit...

larskanis:
> relatively easy solution that even 'never touched windows' types could implement

Really? 700 lines of powershell code? I wouldn't want to maintain such a mess in my repository. (Having a distaste for powershell...)

Although I still think that fat binary gems often [do more harm than good](https://github.com/codahale/bcrypt-ruby/issues/173), they are required to support Ruby older than 2.4, if the gem has external dependencies. This is the case for sqlite3. So I think we shouldn't drop binary versions of sqlite3-ruby now.

Yes, Appveyor can be a solution for compiling windows gems, but [rake-compiler-dock](https://github.com/rake-compiler/rake-compiler-dock) is already implemented for sqlite3-ruby. And it is portable enough to run locally on any system. With it a gem author can debug local instead of on a CI runner. So do we need a second option to build fat binary gems?

Anyway, why on hell don't you use ruby to build these rubygems? That could be delivered as gem instead of powershell scripts copied from one to the next repository. It would be much more acceptable for gem maintainers. For comparison: for sqlite3-ruby [this commit](https://github.com/sparklemotion/sqlite3-ruby/commit/c8552f57fdf0e24571dc0ad03f14965853bc767b) added cross compilation, based on the support in rake-compiler. And [this commit](https://github.com/sparklemotion/sqlite3-ruby/commit/fb0e752554a6ce86c1794916306be375425e0b57) added rake-compiler-dock in just 5 lines of code.

But IMHO binary gems are too different to the way gems are installed and used on Linux and MacOS. Actually they make Windows a second class OS, since they require special support and knowledge by the gem author (like timely releases for each new ruby release). And they make it harder to solve Windows specific issues of the given gem, since you can't change the binary part.

In the long term I think it's better to investigate why the installation of particular source gems fail. With MSYS2 we have a pretty homogeneous build environment compared to the many Linux distributions. And with the [Ruby+Devkit flavour](https://rubyinstaller.org/downloads) it's easy as pie to install this working build environment on Windows.

Having maintained a lot of binary gems since years (pg, ffi, libusb, fxruby, pkcs11, nokogiri), I also got various gem specific issues around them as well - beside the generic issues mentioned [at bcrypt](https://github.com/codahale/bcrypt-ruby/issues/173).

For sqlite3-ruby we actually need a new release with binary support for Ruby-2.5. It's all on git master, but not yet released.

MSP-Greg:
Lars,

> Really? 700 lines of powershell code? I wouldn't want to maintain such a mess in my repository. (Having a distaste for powershell...)

Not saying I like PowerShell that much, but it sure beats bat/cmd script.

Really? 700 lines of powershell code?

Three files totaling 128 lines of code.  One is a replacement rake file, one is ruby to assemble the gem, the last is the ps1 file, most of which is for compiling SQLite3.  Puma required 81 lines in the three files.

The files in the shared folder are common to any gem using the framework.  They could be a submodule, but that seems like a lot of trouble for three files.  Also, much of the code formats the output for Appveyor, parses the test results, handles the artifacts, etc.

Remember also that this is building the fat-binary, then installing it in every Ruby version, then testing it.

Finally, It does run locally on a Window system, with all the Ruby versions and build tools installed.  An additional ps1 needs to be added which has all the local paths defined.

Could I write it in Ruby?  Probably.  After completing it in Powershell, I'm not really interested in moving to Ruby unless people are interested in using it.

Regardless of the framework, it certainly appears that many gem owners/maintainers are having problems building Windows fat-binary gems.  I developed this to address that.  Do you consider that to be an issue?

There are several issues here:
1. Should support for fat-binary gems continue?
2. Are current fat-binary build systems meeting the needs of extension gem owners/maintainers, especially the ones unfamiliar with Windows?
3. Depending on the above, should fat-binary gem construction be integrated into the test process on Appveyor?

You and I don't need fat-binary gems.  But I think others do.

Thanks, Greg

MSP-Greg:
Closing due to new revision that retrieves shared folder files from an Appveyor artifact, and hence, the folder is no longer required in the repo.  Makes it simpler for me to revise them also.

PR #237, opened 2018-06-21
Build Windows fat-binary gems on Appveyor

MSP-Greg:
Changes Appveyor CI to allow fat-binary gem construction along with testing.

Also patches extconf.rb since SQLite appears to account for dl functions when compiled under windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,9 @@ ports
 InstalledFiles
 Makefile
 Gemfile.lock
+
+# windows local build artifacts
+win_gem_test/shared/
+win_gem_test/test_logs/
+/Rakefile_wintest
+*.gem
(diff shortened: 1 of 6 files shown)


MSP-Greg:
See https://ci.appveyor.com/project/MSP-Greg/sqlite3-ruby/build/3

MSP-Greg:
Details not important, but I revisited this.  Current Appveyor build is at:
https://ci.appveyor.com/project/MSP-Greg/sqlite3-ruby

All tests pass for 32 & 64 bit, Ruby 2.0 thru 2.6, using SQLite3 3.27.1.

I forgot, but the build is static, don't have dependency on any external library/dll.  Does make the pre-compiled gem larger...

PR #251, opened 2019-02-18
travis.yml - fix gem update --system

MSP-Greg:
RubyGems is updated in travis.yml.

With the release of RubyGems 3.0, it's own ignorance of `required_ruby_version` causes it to not be able to update itself.  PR fixes that by adding conditional script code.

Since both RubyGems 2.7.8 & current versions include Bundler, commented out its update.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,16 @@
 sudo: false
 language: ruby
 cache: bundler
+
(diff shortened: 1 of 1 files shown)


MSP-Greg:
Closing due to CI updates/changes

PR #253, opened 2019-03-05
Fix required ruby version in gem

jeremyevans:
The gem no longer builds on 1.8.7 as it requires encoding supported added in ruby 1.9.


jeremyevans:
Travis-CI failures are due to rubygems-update not being supported for ruby versions <2.3, unrelated to this change.

flavorjones:
Thanks for submitting this change, @jeremyevans. I've just started a branch to overhaul the gemspec, the CI matrix, and how the native gems are packaged, and I'll pull this commit into that changeset.

Issue #250, opened 2019-02-15
Version 1.4 broken on Windows


sroller:
Environment: Windows 10, msys2, ruby 2.5.3p105 x64

When I install the gem version 1.4 the installation fails with a link error: can't find dlopen().
I reverted to version 1.3.11 of the gem and things where fine.
Let me know what additional information you need.

MSP-Greg:
Try adding the mingw-w64-x86_64-dlfcn package, then running `gem install`.

I think:
```
ridk exec pacman -S mingw-w64-x86_64-dlfcn
```

sroller:
Yes, that worked. We should change the comment in the README.md of the rubyinstaller2 project to reflect on this change.

MSP-Greg:
Glad that helped.  I think that 'dlfcn' should be added to:

https://github.com/sparklemotion/sqlite3-ruby/blob/5c36dbeb8f61a13423d233380047d8051b7eefb7/sqlite3.gemspec#L10

and/or the following should be modified when building on Windows:

https://github.com/sparklemotion/sqlite3-ruby/blob/5c36dbeb8f61a13423d233380047d8051b7eefb7/ext/sqlite3/extconf.rb#L67-L71

haven't had a chance to look into it...

PR #260, opened 2019-03-27
Add dlfcn as dependency for MSYS2

### LINKED ISSUE #259: sqlite3 1.4.0 does not build on Windows ("missing function dlopen")
`gem install` fails on Windows with error `missing function dlopen`.

Steps (in my case on Windows 7 Professional, 64 Bit, Ruby 2.5.1, msys2):

- install Ruby via rubyinstaller-2.5.1-2-x64
- `ridk install`
- `gem update —system` _(unrelated to the problem, necessary for bundler)_
- `gem install sqlite3 -v 1.4.0`

After some hours of searching, testing, and frustration I came up with the following workarounds that work for my setting:

## Fix 1

`bundle install` with a Gemfile that includes the following line:

```
gem "sqlite3", git: "https://github.com/larskanis/sqlite3-ruby", branch: "revert-dl"
```

This reverts the check for the presence of the dl functions, as proposed in #254.

In my case the install succeeded, and I was able to use sqlite3 for accessing a database. I have no idea whether these functions are needed for some functionality, and whether this might blow up at some point. The mentioned PR states they are not needed.

(This is similar to a fix that can be found in various issues and blog posts, there with `branch: "add-gemspec"`, but with an updated branch based on current master.)

## Fix 2a

Manually install the missing dl functions before trying to install the gem, via

```
$ ridk exec pacman -S mingw-w64-x86_64-dlfcn
```

As suggested in #250.

## Fix 2b

`bundle install` with a Gemfile that includes the following line:

```
gem "sqlite3", git: "https://github.com/stomar/sqlite3-ruby", branch: "add-dlfcn"
```

This is similar to fix 2a, but instead of the manual install adds `dlfcn` as a dependency to the slite3 gemspec, see #260.

## Conclusion

Please consider to either

- merge #254, in the case that dlfcn is indeed not needed,
- or include dlfcn as a dependency (#260), so that a manual install is not necessary.

Note that there already have been two issues opened for this problem (#250, #248), but they have been closed (prematurely IMO) by the authors after they had been provided with a workaround.

stomar:
Add the "dlfcn" library as MSYS2 dependency for RubyInstaller2.
This fixes the "missing function dlopen" error during installation.
(Introduced with f4ffec281a2888c1e536662ac1ea740fd3f5433c, cc @tenderlove.)

This fixes #259 and is an alternative solution to #254, assuming that dlfcn _is_ necessary for sqlite3.

This essentially has been suggested by @MSP-Greg in #250.

I need to say that I have no idea whether adding this dependency might have some undesired side effects.

_Also note that the gemspec is outdated (wrong version, dates, ...). I regenerated with `rake gem:spec`, but only committed the relevant line._

### DIFF
--- a/rakelib/gem.rake
+++ b/rakelib/gem.rake
@@ -23,7 +23,7 @@ HOE = Hoe.spec 'sqlite3' do
   require_rubygems_version ">= 1.3.5"
 
   spec_extras[:extensions] = ["ext/sqlite3/extconf.rb"]
-  spec_extras[:metadata] = {'msys2_mingw_dependencies' => 'sqlite3'}
+  spec_extras[:metadata] = {'msys2_mingw_dependencies' => 'sqlite3 dlfcn'}
 
   extra_dev_deps << ['rake-compiler', "~> 1.0"]
   extra_dev_deps << ['rake-compiler-dock', "~> 0.6.0"]
--- a/sqlite3.gemspec
+++ b/sqlite3.gemspec
@@ -7,7 +7,7 @@ Gem::Specification.new do |s|
   s.version = "1.3.13.20180326210955"
 
   s.required_rubygems_version = Gem::Requirement.new(">= 1.3.5".freeze) if s.respond_to? :required_rubygems_version=
-  s.metadata = { "msys2_mingw_dependencies" => "sqlite3" } if s.respond_to? :metadata=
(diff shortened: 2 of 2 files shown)


flavorjones:
I believe that the merger of #254 addresses this issue.

PR #277, opened 2019-12-18
Remove deprecated taint followup

casperisfine:
This is a followup on @kamipo's #273 

It also avoid calling `rb_check_safe_obj` on MRI 2.7. That warning is totally spamming Active Record's test suite.

cc @tenderlove since apparently you're maintainer.

cc @rafaelfranca @Edouard-Chin.

NB: Based on @kamipo's PR I expect half of the CI jobs to fail, I'll try to make them pass by pushing another commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: ruby
 cache: bundler
 before_install:
-  - gem update --system
-  - gem install bundler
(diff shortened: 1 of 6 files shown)


tenderlove:
Thank you!

kbrock:
Thank you

PR #278, opened 2019-12-20
Travis fix

MSP-Greg:
This shows that CI is working for the current master.  The main fix is in .travis.yml.

The fix in the gemfile is due to hoe not having a `required_ruby_version` in its gemspec, as the last two releases are incompatible with Ruby 1.9.3 & 2.0. I believe it cannot be added to the code for `rake bundler:gemfile`, so it's a temp fix.

Opened an issue in hoe for it this morning.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,14 @@
 language: ruby
 cache: bundler
 before_install:
-  - gem update --system 2.7.7
-  - gem install bundler -v 1.16.2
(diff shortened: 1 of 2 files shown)


MSP-Greg:
@tenderlove

Sorry about the ping in actions, but things were kind of not moving, but they are now.  There's changes coming for Ruby in Actions, but maybe leave things as is for now... (?)

PR #279, opened 2019-12-20
Add GitHub Actions for Windows CI

MSP-Greg:
1. Add rake compile:msys2 to allow compiling with MSYS2 sqlite3 package

2. Add .github/workflows/windows.yml to run GitHub Actions on Ruby 2.4 thru master on Windows


MSP-Greg:
MSYS2 is using sqlite 3.30.1.  There are compile warnings and the tests have 'deprecated' warnings.

PR #283, opened 2020-03-16
Handle the file argument properly

yb66:
The argument is a file *path* but is called *file* and yet it's treated as a string.
If passed an actual File or a Pathname it will be properly handled.

### DIFF
--- a/lib/sqlite3/database.rb
+++ b/lib/sqlite3/database.rb
@@ -65,6 +65,7 @@ module SQLite3
     def initialize file, options = {}, zvfs = nil
       mode = Constants::Open::READWRITE | Constants::Open::CREATE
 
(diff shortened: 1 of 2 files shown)


tenderlove:
@yb66 do you mind adding a test for this?

MSP-Greg:
@yb66

Windows behaves a bit differently with file handles.  The test below passes CI:
```ruby
def test_filename_to_path
  tf = Tempfile.new 'thing'
  pn = Pathname tf.path
  db = SQLite3::Database.new pn
  assert_equal pn.expand_path.to_s, File.expand_path(db.filename)
ensure
  tf.close! if tf
end
```

yb66:
Thanks @MSP-Greg.

Regards,
iain

PR #304, opened 2021-12-08
Fix unit tests

orgads:


dbungert:
While this will make the tests work for sqlite 3.37.0+, it breaks them pre-3.37.  For compatibility with both, it seems necessary to declare these types in uppercase, or do a case insensitive compare.

Relevant sqlite commit:
https://github.com/sqlite/sqlite/commit/c2df4d6adb44e5f9587bf962c917c46c603825c9

Another package with a similar problem:
https://github.com/DBD-SQLite/DBD-SQLite/commit/ba4f472e7372dbf453444c7764d1c342e7af12b8

flavorjones:
Fixed in #315

PR #315, opened 2022-03-18
Add Ruby 3.0 and Ruby 3.1 to the CI matrix

petergoldstein:
In addition to updating the CI configuration to include Ruby 3.0 and Ruby 3.1, a number of other changes were necessary to get this to green across the board.  I could use a review by the maintainers to ensure these were done correctly.  They were:

1. Updating from the use of the unsupported `mini_portile` gem to the `mini_portile2` gem, as the former isn't compatible with Ruby 3.1.
2. Case normalizing the type values returned by `table_info` to lower case, which is what the tests expect.  On Mac and Windows these were returned upper case (this test failure exists in CI before this PR)
3. Updating a test that directly accesses the types of a query result to accept both upper and lower case values

With these changes, everything runs green.


casperisfine:
👍 , I just ran into the same thing. cc @tenderlove @flavorjones, would be nice to merge this.

flavorjones:
I've kicked CI to demonstrate green but I've not been involved with sqlite3 in 12 years. I'd like someone who's more familiar with sqlite3 and this code to determine whether the downcase fixes for column types is the right thing to do (since it's not clear to me why or when this behavior changed). @tenderlove ?

test/test_integration_resultset.rb
```diff
@@ -118,7 +118,7 @@ def test_enumerable
   end
 
   def test_types
-    assert_equal [ "integer", "text" ], @result.types
+    assert_equal [ "integer", "text" ], @result.types.map { |t| t.downcase }
```
casperisfine, line 121: Why do we need to downcase here if it's already handled in `pragmas.rb`?

petergoldstein: @casperisfine Because the `pragmas.rb` change only impacts the `table_info` call.  The `types` call here goes directly to the extension.

casperisfine: My bad. Can we add a similar `downcase` in there to preserve backward compatiblity?

petergoldstein: @casperisfine I made that change.  Please take a look and let me know what you think.

test/test_statement.rb
```diff
@@ -127,7 +127,7 @@ def test_bind_blob
       row = @db.execute('select * from foo')
 
       assert_equal ['hello'], row.first
-      assert_equal row.first.types, ['BLOB']
+      assert_equal ['blob'], row.first.types
```
casperisfine, line 130: I think I'm confused now... Some types were capitalized before but not all of them???

petergoldstein: I did this a while back, so frankly I don't remember the details off the top of my head.  But you can see my comments in #2 in my original description.   I was seeing case inconsistency across different operating systems.

petergoldstein: @casperisfine You can see an action that demonstrates the issue here - https://github.com/sparklemotion/sqlite3-ruby/runs/6090043254 .  Ubuntu passes (lowercase) but Windows/Mac does not.  This was ultimately all coming from the underlying C code, whose behavior didn't seem to be consistent across operating systems.  So I had the Ruby code case normalize, based on what had been in the specs (assuming that was the historically expected behavior)

casperisfine: Ok, thanks for the info, I'll try to dig into this a bit more next week. We also need to look at what the common callers expects (e.g. Active Record)

petergoldstein: @casperisfine Looking it ActiveRecord's type matching, it looks like it's explicitly case insensitive - https://github.com/rails/rails/blob/74ba52ec5c5aa53e2e728d88c179b2d487fff2b4/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L736

casperisfine:
@tenderlove / @flavorjones I suggest we merge this.

flavorjones:
@casperisfine do you need a release, too?

casperisfine:
@flavorjones not yet, I wanted this merged to fix CI on https://github.com/sparklemotion/sqlite3-ruby/pull/317, for which I'd love a release yes.

PR #317, opened 2022-05-15
Add a `strict: true` option.

casperisfine:
Add a strict: true option.
Ref: https://github.com/rails/rails/pull/45101
Ref: https://github.com/rails/rails/issues/27782
Ref: https://www.sqlite.org/quirks.html#double_quoted_string_literals_are_accepted

Sqlite has a really unfortunate behavior around double quotes. It first tries to consider them as column names, but if they don't exist it then consider them as string literals.

Because of this, typos can silently go unnoticed. It would be very useful for Active Record if this behavior could be disabled.

cc @tenderlove @fatkodima

### DIFF
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -66,6 +66,23 @@ static VALUE rb_sqlite3_open_v2(VALUE self, VALUE file, VALUE mode, VALUE zvfs)
   return self;
 }
 
+static VALUE rb_sqlite3_disable_quirk_mode(VALUE self)
+{
+#if defined SQLITE_DBCONFIG_DQS_DDL
+  sqlite3RubyPtr ctx;
+  Data_Get_Struct(self, sqlite3Ruby, ctx);
+
+  if(!ctx->db) return Qfalse;
+
+  sqlite3_db_config(ctx->db, SQLITE_DBCONFIG_DQS_DDL, 0, (void*)0);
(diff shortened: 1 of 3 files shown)


ext/sqlite3/database.c
```diff
@@ -66,6 +66,23 @@ static VALUE rb_sqlite3_open_v2(VALUE self, VALUE file, VALUE mode, VALUE zvfs)
...
+  Data_Get_Struct(self, sqlite3Ruby, ctx);
+
+  if(!ctx->db) return Qfalse;
+
+  sqlite3_db_config(ctx->db, SQLITE_DBCONFIG_DQS_DDL, 0, (void*)0);
+  sqlite3_db_config(ctx->db, SQLITE_DBCONFIG_DQS_DML, 0, (void*)0);
+
+  return Qtrue;
+#else
+  return Qfalse
```
hurricup, line 82: Feels like both branches should be covered with at least building tests to avoid such stuff :(

casperisfine: Avoid such what?

Also I really don't see what there is to cover, either the function exist and we use it, or either it's a noop.

hurricup: Avoid compilation error because of missing semicolon after Qfalse.

hurricup: https://github.com/sparklemotion/sqlite3-ruby/pull/324

casperisfine: 🤦 

Getting both versions of sqlite3 on CI is tricky, but we can probably have an ENV var flag.

PR #324, opened 2022-06-13
Add missing semicolon

r6e:
Fixes #323 

The build fails on the `ruby:2.7.5-buster` Docker image (and, presumably, other Linux distros using the same build chain) due to a missing semicolon.  For some reason, the build does _not_ fail on macOS.

Adding this semicolon allows the build to pass on the `ruby:2.7.5-buster` image, and it still passes on macOS as well.

PR #320, opened 2022-05-25
native gems, better system library support, more platform support

flavorjones:
Much time has passed since the last time we shipped precompiled native gems for sqlite3, and we need to do some work to get them to build now.

This branch is a work-in-progress which will close #301 and close #313

- [x] remove hoe
- [x] upgrade sqlite3 to 3.38.5
- [x] get linux 2.5..3.1 green
- [x] get macos 2.5..3.1 green
- [x] get windows 2.5..3.1 green
  - [x] #255
  - [x] #231 
  - [x] #259 
  - [x] get appveyor green: https://ci.appveyor.com/project/flavorjones/sqlite3-ruby
  - [x] mswin support (from #321)
- [x] look into M1 support
  - [x] #313 
  - https://sqlite.org/forum/info/73165e901d798571
- [x] test system libs with sqlite3
- [x] test system libs with sqlcipher
- [x] ~look into using https://rubygems.org/gems/native-package-installer/ for system library users~
- [x] test native gems across all platforms
- [x] make sure all the gems package up the right things:
  - [x] native gems should not include tarball or patches, but should still include C files for rdoc processing
  - [x] source gem should include tarball and patches
  - [x] mini_portile2 should not be in the native gem specs
  - [x] gemspec metadata for `msys2_mingw_dependencies` should not be present for native gems
- [x] document installation options:
  - [x] native gems
  - [x] building from source against the packaged tarball
  - [x] building from source against system sqlite3
  - [x] building from source against system sqlcipher
  - [x] supported versions of ruby
  - [x] supported versions of sqlite3
- [x] update license information

PR #326, opened 2022-06-14
version bump to v1.4.4

flavorjones:
Let's please ship a new version to fix an issue (#324) introduced in v1.4.3.


flavorjones:
Please note that the appveyor failures can be ignored for now, I've fixed them in #320 which isn't quite ready for review yet.

nvolker (approved):
Would appreciate this going in. Currently breaking CI and I'd prefer not locking down to 1.4.2. Thanks for doing this 👍

PR #328, opened 2022-07-05
dep: upgrade sqlite to 3.39.0

flavorjones:
sqlite 3.39.0 was released last week.

This PR updates to that version, and removes the previous patch needed to compile on arm64-darwin.

See https://sqlite.org/releaselog/3_39_0.html for release notes.


flavorjones:
Looks like some tests are failing on darwin related to symbolic links, likely due to this entry in the sqlite changelog:

> The unix os interface resolves all symbolic links in database filenames to create a canonical name for the database before the file is opened.

PR #333, opened 2022-08-09
test: assert compiled and loaded versions of sqlite3 are the same

flavorjones:
Let's assert on this to make sure that CI is testing what we think it's testing.

I have reason to believe (as part of an investigation into #332) that on macos using system libraries, something funky is going on. And it's not obvious to me what it is.


flavorjones:
Yeah, macos failed as expected. Not sure how to address this, so I'll leave it here for now for anyone who has the time to dig into a diagnosis.

flavorjones:
#336 supersedes this.

PR #334, opened 2022-08-11
Fix utf16_string_value_ptr() for TruffleRuby

eregon:
* TruffleRuby uses TruffleString which requires all UTF-16 strings to
  always have an even bytesize.
* The code here ensures the UTF-16 string always ends up with 2 \x00
  bytes, so append them explicitly to have an even bytesize instead of
  relying on RSTRING_PTR() adding the second \x00 byte.
* Fixes https://github.com/oracle/truffleruby/issues/2704

This fixes the CI on TruffleRuby.

### DIFF
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -31,7 +31,7 @@ static char *
 utf16_string_value_ptr(VALUE str)
 {
   StringValue(str);
-  rb_str_buf_cat(str, "\x00", 1L);
+  rb_str_buf_cat(str, "\x00\x00", 2L);
   return RSTRING_PTR(str);
 }
 


eregon:
Ah, the side effect of `sqlite3_open16` is to set the database encoding to UTF-16 instead of UTF-8 (https://www.sqlite.org/c3ref/open.html).
And I'm not sure there is a way to set "the default encoding for databases" after `sqlite3_open_v2`.
So this change makes most sense and we can't remove `sqlite3_open16` too easily if users want a UTF-16 "default encoding for databases".

eregon:
FWIW CRuby master seems to have [`TERM_LEN`](https://github.com/ruby/ruby/blob/8a3f401b24e1b5f3e3a0f44c568a3e66ed595d42/string.c#L167) logic to ensure there are `rb_enc_mbminlen()` `\x00` bytes at the end, but that's currently not implemented on TruffleRuby, and I'm not sure since which version of CRuby that's done.

eregon:
> but that's currently not implemented on TruffleRuby

That seems easy to fix, so I have an incoming fix for that in TruffleRuby.
So if we're happy to rely on Ruby having enough `\x00` bytes, then we can replace `utf16_string_value_ptr()` with `RSTRING_PTR()`.
But CRuby 1.9.2 does not seem to have the TERM_LEN logic, e.g. https://github.com/ruby/ruby/blob/ruby_1_9_2/string.c#L410

flavorjones:
@eregon Can you please add a test for this behavior?

flavorjones:
Ah, I see, it fails when truffleruby uses packaged libraries, but not when it uses system libraries. No worries, I'll work through it.

flavorjones:
Thank you!

PR #336, opened 2022-08-12
test: assert compiled and loaded versions of sqlite3 are the same, update sqlite3-ruby.yml

MSP-Greg:
Similar to PR #333, adds a test to verify that compile & runtime sqlite3 versions are the same.

From watching Puma & EventMachine (both compile with OpenSSL), macOS has issues when using brew packages, the solution here is to use `--with-opt-dir=$(brew --prefix sqlite3)` when compiling.

Third comment can be removed, it just contains reformatting of the Actions yaml...

Failure with TruffleRuby-head, see PR #334 for info.  It is a recent failure.

### DIFF
--- a/.github/workflows/sqlite3-ruby.yml
+++ b/.github/workflows/sqlite3-ruby.yml
@@ -1,3 +1,5 @@
+# for Actions OS info, see https://github.com/actions/runner-images#available-images
+
 name: test suite
 concurrency:
   group: "${{github.workflow}}-${{github.ref}}"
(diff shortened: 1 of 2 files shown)


flavorjones:
Thank you for this! Merging.

MSP-Greg:
Mike,

I was working a fix for the other workflow file, as I didn't notice failures in it before when looking at PR #333.

I also had a message, but was doing both at the same time you merged this.  I open another PR with a fix.

Issue #335, opened 2022-08-12
ver1.5.0.rc1, execute method is 30% slower than ver1.4.4


Watson1978:
Seems that `execute` method is 30% slower than ver1.4.4.

## Environment
- macOS
  - macOS 13.0 Beta（22A5321d）
  - Apple M1 Max
  - Apple clang version 14.0.0 (clang-1400.0.29.102)
  - ruby 3.2.0dev (2022-08-09T11:11:48Z master ff07e5c264) [arm64-darwin22]

## 1.5.0.rc1 result
```
*** SQLite3 1.5.0.rc1 ***
Warming up --------------------------------------
             execute     6.460k i/100ms
Calculating -------------------------------------
             execute     64.732k (± 0.3%) i/s -    329.460k in   5.089641s
```

## 1.4.4 result
```
*** SQLite3 1.4.4 ***
Warming up --------------------------------------
             execute     8.345k i/100ms
Calculating -------------------------------------
             execute     83.406k (± 0.5%) i/s -    417.250k in   5.002748s
```

## Test code
```ruby
require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'benchmark-ips'
  gem 'sqlite3'
end

puts "*** SQLite3 #{SQLite3::VERSION} ***"

db = SQLite3::Database.new(':memory:')
db.execute(<<~SQL)
  create table users (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    name TEXT,
    email TEXT,
    created_at TIMESTAMP,
    updated_at TIMESTAMP
  )
SQL

5.times do |i|
  db.execute("insert into users (name, email, created_at, updated_at) values ('foo', 'foo@example.com', '#{Time.now.strftime("%Y-%m-%d %H:%M:%S")}', '#{Time.now.strftime("%Y-%m-%d %H:%M:%S")}')")
end


Benchmark.ips do |x|
  # x.time = 20

  query = 'select * from users'
  x.report('execute') {
    db.execute(query)
  }
end
```

flavorjones:
Thanks for opening this issue, I will spend some time to try to reproduce and investigate before we cut a 1.5.0 final.

flavorjones:
Hi! Thanks for your patience. Because this gem previously used system libraries, benchmark results are going to heavily depend on what version of sqlite3 you're using and how it was built.

[...]

PR #341, opened 2022-08-29
perf: use -O2 compiler optimizations

flavorjones:
@Watson1978 pointed out in #335 that the 1.5.0.rc1 performance was worse than 1.4.4. At least part of the reason for this is that we didn't turn on compiler optimizations in 1.5.0.rc1.

This should improve the situation for both the precompiled native and the vanilla gems.

### DIFF
--- a/ext/sqlite3/extconf.rb
+++ b/ext/sqlite3/extconf.rb
@@ -51,7 +51,11 @@ module Sqlite3
         minimal_recipe.tap do |recipe|
           recipe.configure_options += ["--enable-shared=no", "--enable-static=yes"]
           ENV.to_h.tap do |env|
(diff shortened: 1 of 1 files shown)


Watson1978:
Seems that my issue will be solved by this PR. Thanks

```
$ ruby bench.rb
info: sqlite3-ruby version: 1.5.0.rc1/1.5.0.rc1
info: sqlite3 version: 3.39.2
info: sqlcipher?: false
info: threadsafe?: true
Warming up --------------------------------------
             execute     8.772k i/100ms
Calculating -------------------------------------
             execute     87.746k (± 0.5%) i/s -    447.372k in   5.098649s
```

flavorjones:
@Watson1978 Excellent! Thank you *so much* for trying out the RC and for reporting this.

sneak-hug

PR #345, opened 2022-09-07
rb_gc_register_address() must be called after the variable was assigned

eregon:
* Otherwise the GC might read the variable when it is not pointing to a Ruby object yet.
* TruffleRuby also assumes it can read the pointer when
  rb_gc_register_address() is called: https://github.com/oracle/truffleruby/issues/2720

### DIFF
--- a/ext/sqlite3/aggregator.c
+++ b/ext/sqlite3/aggregator.c
@@ -269,6 +269,6 @@ rb_sqlite3_aggregator_init(void)
   cAggregatorWrapper = rb_funcall(rb_cClass, rb_intern("new"), 0);
   cAggregatorInstance = rb_funcall(rb_cClass, rb_intern("new"), 0);
 
-  rb_gc_register_address(&cAggregatorWrapper);
-  rb_gc_register_address(&cAggregatorInstance);
+  rb_gc_register_mark_object(cAggregatorWrapper);
+  rb_gc_register_mark_object(cAggregatorInstance);
 }
--- a/ext/sqlite3/aggregator.c
+++ b/ext/sqlite3/aggregator.c
@@ -265,9 +265,10 @@ rb_sqlite3_define_aggregator2(VALUE self, VALUE aggregator, VALUE ruby_name)
 void
 rb_sqlite3_aggregator_init(void)
 {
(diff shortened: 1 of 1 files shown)


flavorjones (changes requested):
My understanding is that compaction _could_ occur at any time if we re-enter the Ruby interpreter. If that's true, I think this should be:

```patch
diff --git a/ext/sqlite3/aggregator.c b/ext/sqlite3/aggregator.c
index 7b07f17..42044bb 100644
--- a/ext/sqlite3/aggregator.c
+++ b/ext/sqlite3/aggregator.c
@@ -265,9 +265,10 @@ rb_sqlite3_define_aggregator2(VALUE self, VALUE aggregator, VALUE ruby_name)
 void
 rb_sqlite3_aggregator_init(void)
 {
-  rb_gc_register_address(&cAggregatorWrapper);
-  rb_gc_register_address(&cAggregatorInstance);
   /* rb_class_new generatos class with undefined allocator in ruby 1.9 */
   cAggregatorWrapper = rb_funcall(rb_cClass, rb_intern("new"), 0);
+  rb_gc_register_mark_object(cAggregatorWrapper);
+
   cAggregatorInstance = rb_funcall(rb_cClass, rb_intern("new"), 0);
+  rb_gc_register_mark_object(cAggregatorInstance);
 }
```

WDYT? Am I misunderstanding the intent of `rb_gc_register_*`?

eregon:
I pushed a commit to change to that.

flavorjones:
Thanks again!

eregon:
In the light of https://bugs.ruby-lang.org/issues/19584, it would make sense to revert this.
However, https://github.com/oracle/truffleruby/issues/2721 is only fixed in 23.0 (not in 22.3), so it would be best to wait until TruffleRuby 23.0 is released.

flavorjones:
I'd prefer not to revert, but I did see the upstream fix and am happy this won't be a problem for other C extensions.

eregon:
OK, that's the simplest then, thanks.
It's probably a very very unlikely issue on CRuby and also it will be fixed for the next set of CRuby releases.

PR #347, opened 2022-09-11
ci: install gem under truffleruby

flavorjones:
Trying to reproduce https://github.com/oracle/truffleruby/issues/2729


eregon:
It should work with a rerun: https://github.com/oracle/truffleruby/issues/2729#issuecomment-1274679978

And thank you for adding truffleruby in CI :)

Issue #299, opened 2021-05-03
Ractor support


sandlerr:
Hello maintainers,

Is there any interest or planned change to add support to this gem for use within a Ruby 3 [Ractor](https://docs.ruby-lang.org/en/master/doc/ractor_md.html)?

Currently, constructing a database instance is failing with the error message introduced into [this Ruby commit](https://git.ruby-lang.org/ruby.git/commit/?id=182fb73c40). I have collected this stacktrace:

```
~/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/sqlite3-1.4.2/lib/sqlite3/database.rb:89:in `open_v2': ractor unsafe method called from not main ractor (Ractor::UnsafeError)
        from ~/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/sqlite3-1.4.2/lib/sqlite3/database.rb:89:in `initialize'
        ...
```

It can be reproduced by:

```ruby
sqlite_ractor = Ractor.new do
    db = SQLite3::Database.open "xyz.db"
end
puts sqlite_ractor.take
```

The Ruby commit message justifies the error because:

> C extensions can violate the ractor-safety

However, this gem doesn't appear to store references to Ruby objects that might be accessed across Ractors / invalidly shared, so perhaps its C extensions can be marked as Ractor-safe?

tenderlove:
Hi,

I would like to support Ractors, but I'm not sure how we would do that exactly.  SQLite3 is a C library, so we must make calls to C functions and as such this is a Ruby C extension.  If there is a way to make C extensions play nicely with Ractors, then I'm all for it! 😄

sandlerr:
@tenderlove , thank you for your feedback from last year. It can sort of be done, but unfortunately not yet to the point where one could conceivably build something like a connection pool - not that this is a useful thing for SQLite users, but it could be a starting point for other projects exploring ractors.

Basic support and testing was built by @kjtsanaktsidis and I've included it with commented details on the limitations being tested here: https://github.com/sparklemotion/sqlite3-ruby/compare/master...sandlerr:sqlite3-ruby:sandlerr/ractor

>  If there is a way to make C extensions play nicely with Ractors, then I'm all for it!

This part is a single C API call (after eyeballing some code and not seeing obvious violations of the ractor rules, like storing `VALUE`s that might be accessed incorrectly in another thread).

tenderlove:
[...]

PR #365, opened 2022-12-28
Initial Ractor support

sandlerr:
Allows users to create / open databases, send queries, and process ResultSets inside a Ractor. However, does not allow passing the gem's domain objects through the Ractor communication channels.

See https://github.com/sparklemotion/sqlite3-ruby/issues/299 .


test/test_integration_ractor.rb
```diff
@@ -0,0 +1,84 @@
...
+  def setup
+    teardown
+  end
+
+  def teardown
+    FileUtils.rm_rf(Dir.glob "#{STRESS_DB_NAME}*")
+  end
+
+  def test_ractor_share_database
+    skip('Requires Ruby with Ractors') if RUBY_VERSION < '3.2.0'
```
flavorjones, line 18: Can I ask that you add a predicate method on the `SQLite3` module, similar to `.threadsafe` and `.sqlcipher?` that can be used self-describe and to skip tests when necessary? It's much more descriptive than a version number check.

flavorjones (changes requested):

test/test_integration_ractor.rb
flavorjones: (And print out the value of that predicate in `test/helper.rb` where we print the other info.)

sandlerr:
Is there any outstanding feedback here? Thanks

tenderlove (approved):
This seems fine to me. I haven't investigated everything we need to do to make C extensions Ractor safe but I think this is fine to start with!

flavorjones:
Great, I'll rebase and merge it today or tomorrow.

flavorjones (approved):

flavorjones:
So it seems like in Ruby 3.3, trying to pass a `Database` between ractors results in a `Ractor::RemoteException` being raised; but in earlier versions of Ruby it makes an implicit copy of the object. This means that the `test_ractor_share_database` test hangs.

I'm not sure if this is intentional behavior or not. @sandlerr or @tenderlove do you know?

flavorjones:
OK, that exception started to get raised in https://github.com/ruby/ruby/commit/ce47ee00 as a fix, so I'm just going to adapt the test to the different behaviors before and after 3.3.0.

flavorjones:
The stress test hangs under Ruby 3.3 with as few as 3 ractor writers, and it's not clear to me why.

tenderlove:
I made some updates to this branch and included a [RACTOR.md](https://github.com/sandlerr/sqlite3-ruby/blob/sandlerr/ractor/RACTOR.md) file that describes Ractor support.  Also I removed the "Ractor smoke test" test because I think it's just testing the guarantees that the SQLite library makes and not testing _our_ APIs.

@flavorjones please take a look at the `RACTOR.md` file, I think it covers most of the information behind Ractor support.  I _think_ this PR is good to merge for the most part.

flavorjones:
Rebased onto `main`

flavorjones:
Looks like the FreeBSD build is failing:

- https://github.com/sparklemotion/sqlite3-ruby/actions/runs/14157606581/job/39658900400?pr=365#step:3:2126

```
/home/runner/work/sqlite3-ruby/sqlite3-ruby/test/test_collation.rb:64: [BUG] pthread_mutex_lock: Resource deadlock avoided (EDEADLK)
  ruby 3.2.6 (2024-10-30 revision 63aeb018eb) [amd64-freebsd14]
```

flavorjones:
@tenderlove It looks like there are still problems with this branch. Looks like a few other tests have deadlocked or are hanging, in addition to FreeBSD.

flavorjones:
Also, I don't see RACTOR.md in this branch?

flavorjones:
I may have rebased this incorrectly ... not sure what's going on but when I rebase RACTOR.md disappears. I'll come back to it later today.

tenderlove:
> I may have rebased this incorrectly ... not sure what's going on but when I rebase RACTOR.md disappears. I'll come back to it later today.

I think I accidentally committed to the merge commit. I can fix up the commits (but it'll be a while)

tenderlove:
> @tenderlove It looks like there are still problems with this branch. Looks like a few other tests have deadlocked or are hanging, in addition to FreeBSD.

I was going to ask if we do any forking in the tests, but we have the fork tracker, so clearly we do.  IIRC there are problems with Ractors and forking such that we can't create a Ractor _and_ fork in the same process (at the moment, it's a bug we're trying to fix).  I think we could either break the Ractor tests in to a different suite (maybe something like `rake test:ractors`) or maybe run them in a sub-process?

 cc @jhawthorn

flavorjones:
@tenderlove I've restored your branch at 36af661b, and I was definitely having problems because `git rebase` drops merge commits by default.

I added a commit to fix standardrb errors.

I would like this branch rebased cleanly before we merge it, though.

Letting CI run again.

eregon:
Would love to see this finished/merged, as it makes a huge difference performance-wise: https://bsky.app/profile/intertwingly.net/post/3mw4yxhkgnk2j
I had the impression this gem's extension is already marked Ractor-safe but not yet :sweat_smile:

ext/sqlite3/extconf.rb
```diff
@@ -125,6 +125,9 @@ def configure_extension
         # Functions defined in 2.1 but not 2.0
         have_func("rb_integer_pack")
 
+        # Functions defined in 3.0 but not 2.7
+        have_func("rb_ext_ractor_safe")
+
```
eregon: `HAVE_RB_EXT_RACTOR_SAFE` is defined directly by Ruby headers so there is no need for this `have_func`: https://github.com/ruby/ruby/blob/e77b7b6f1e3759c849d0bd11106e8acaaac0a725/include/ruby/internal/intern/load.h#L239-L249

PR #367, opened 2022-12-31
native gem support: add 3.2, drop 2.6

### LINKED ISSUE #364: ship precompiled support for Ruby 3.2
This is an issue just to let folks know that this is on its way. Some last-minute changes were made in Ruby around how C extensions are linked on MacOS (Darwin) and we're working through the details.

If you're interested, you can read up a bit on those changes at https://github.com/rake-compiler/rake-compiler-dock/issues/87 or https://github.com/sparklemotion/nokogiri/pull/2732

I hope to have a version shipped within the next few days.

flavorjones:
Closes #364

### DIFF
--- a/.github/workflows/gem-install.yml
+++ b/.github/workflows/gem-install.yml
@@ -56,7 +56,7 @@ jobs:
       fail-fast: false
       matrix:
         sys: ["enable", "disable"]
-        ruby: ["2.6", "2.7", "3.0", "3.1", "head", "truffleruby-head"]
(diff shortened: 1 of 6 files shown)


flavorjones:
OK, precompiled gems on Darwin with Ruby 3.2 are suffering from the same problem as Nokogiri, with the Ruby process having already mapped `/usr/lib/sqlite3.dylib` and so:

```
$ ruby -r sqlite3 -e 'puts SQLite3::SQLITE_VERSION, SQLite3::SQLITE_LOADED_VERSION'
3.40.1
3.39.5
```

Going to have to do the `-load_hidden` workaround here, too.

PR #371, opened 2023-01-02
ext: export only Init_sqlite3_native

flavorjones:
This works around new Darwin Ruby 3.2 symbol resolution issues, but also follows best practices around shipping shared libraries.

The result is that the only public (exported) symbol in the extension is `Init_sqlite3_native`, regardless of what platform or version of Ruby you're on.

See https://github.com/rake-compiler/rake-compiler-dock/issues/87 for lots of context.

### DIFF
--- a/ext/sqlite3/extconf.rb
+++ b/ext/sqlite3/extconf.rb
@@ -55,6 +55,7 @@ module Sqlite3
             additional_cflags = [
               "-fPIC", # needed for linking the static library into a shared library
               "-O2", # see https://github.com/sparklemotion/sqlite3-ruby/issues/335 for some benchmarks
+              "-fvisibility=hidden", # see https://github.com/rake-compiler/rake-compiler-dock/issues/87
             ]
             env["CFLAGS"] = [env["CFLAGS"], additional_cflags].flatten.join(" ")
             recipe.configure_options += env.select { |k,v| ENV_ALLOWLIST.include?(k) }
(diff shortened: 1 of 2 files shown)


flavorjones:
See related proposal for Nokogiri at https://github.com/sparklemotion/nokogiri/discussions/2746

tenderlove (approved):

byroot:
> The result is that the only public (exported) symbol in the extension is `Init_sqlite3_native`, regardless of what platform or version of Ruby you're on.

What about `ruby_abi_version` for dev versions of Ruby?

flavorjones:
@byroot I don't actually know much about how that symbol gets used, can you point me at docs to learn more?

C extensions on Windows have always _only_ exported the target function (`Init_*`) so although the ABI version maybe should _also_ be exported, that's something that's already broken on at least one platform. (If I'm mistaken, please let me know.)

I'm planning to propose something upstream (in Ruby) to formalize or automate exposing minimal symbols, since I think most symbols only get exported on Linux and Darwin by accident (it hasn't been an intentional decision to export them) and this recently caused problems with the RBS gem. I'll make sure to include `ruby_abi_version` in the discussion.

byroot:
It's a symbol Peter added starting in 3.2.0dev to better handle ABI breaks. It was supposed to be exposed in 3.2.0 final too but somehow it didn't happen: https://github.com/ruby/ruby/pull/5474

If needed I'm sure Peter can tell you more.

> I'm planning to propose something upstream (in Ruby) to formalize or automate exposing minimal symbols

Huge 👍 . As you may remember this has caused issues with hiredis etc. I agree C-ext shouldn't expose anything by default. There is very little reason for exposing anything more than the init function.

PR #386, opened 2023-04-21
Exposes the `sqlite3_close_v2` function

clemensp:
I've been running into this error while closing the DB:
`unable to close due to unfinalized statements or unfinished backups`

I've audited the app, and all statements are finalized.

Looking at the sqlite docs, it looks like `sqlite3_close_v2` might actually be the close function to call:
https://www.sqlite.org/c3ref/close.html

> If sqlite3_close_v2() is called with unfinalized prepared statements, unclosed BLOB handlers, and/or unfinished sqlite3_backups, it returns [SQLITE_OK](https://www.sqlite.org/rescode.html#ok) regardless, but instead of deallocating the database connection immediately, it marks the database connection as an unusable "zombie" and makes arrangements to automatically deallocate the database connection after all prepared statements are finalized, all BLOB handles are closed, and all backups have finished. The sqlite3_close_v2() interface is intended for use with host languages that are garbage collected, and where the order in which destructors are called is arbitrary.

The intent of this PR is two-folds:
- To check whether you are open to exposing the `sqlite3_close_v2` function through the sqlite3 adapter.
- To check whether the implementation looks correct. It clears the allocated memory on the ruby side the same way as `sqlite3_rb_close`. I think it's safe, but looking to get your thoughts.
  - I'm not sure if the test is good enough either, as it seems to just check for memory de-allocation.

If the changes look good, I'm also looking at doing a backport to v1.4.4 if you are open to it.

Thank you. :)

### DIFF
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -36,6 +36,7 @@ utf16_string_value_ptr(VALUE str)
 }
 
 static VALUE sqlite3_rb_close(VALUE self);
+static VALUE sqlite3_rb_close_v2(VALUE self);
 
 static VALUE rb_sqlite3_open_v2(VALUE self, VALUE file, VALUE mode, VALUE zvfs)
 {
@@ -95,6 +96,26 @@ static VALUE sqlite3_rb_close(VALUE self)
 
   db = ctx->db;
   CHECK(db, sqlite3_close(ctx->db));
+
+  ctx->db = NULL;
+
+  rb_iv_set(self, "-aggregators", Qnil);
+
+  return self;
+}
+
+/* call-seq: db.close
+ *
+ * Closes this database.
+ */
+static VALUE sqlite3_rb_close_v2(VALUE self)
+{
+  sqlite3RubyPtr ctx;
+  sqlite3 * db;
+  Data_Get_Struct(self, sqlite3Ruby, ctx);
+
+  db = ctx->db;
+  CHECK(db, sqlite3_close_v2(ctx->db));
 
   ctx->db = NULL;
 
@@ -813,6 +834,7 @@ void init_sqlite3_database()
   rb_define_private_method(cSqlite3Database, "open16", rb_sqlite3_open16, 1);
   rb_define_method(cSqlite3Database, "collation", collation, 2);
   rb_define_method(cSqlite3Database, "close", sqlite3_rb_close, 0);
+  rb_define_method(cSqlite3Database, "close_v2", sqlite3_rb_close_v2, 0);
   rb_define_method(cSqlite3Database, "closed?", closed_p, 0);
   rb_define_method(cSqlite3Database, "total_changes", total_changes, 0);
   rb_define_method(cSqlite3Database, "trace", trace, -1);
--- a/sqlite3.gemspec
+++ b/sqlite3.gemspec
@@ -1,10 +1,10 @@
 # -*- encoding: utf-8 -*-
-# stub: sqlite3 1.3.13.20180326210955 ruby lib
+# stub: sqlite3 1.4.4 ruby lib
 # stub: ext/sqlite3/extconf.rb
 
 Gem::Specification.new do |s|
   s.name = "sqlite3".freeze
-  s.version = "1.3.13.20180326210955"
+  s.version = "1.4.4"
 
   s.required_rubygems_version = Gem::Requirement.new(">= 1.3.5".freeze) if s.respond_to? :required_rubygems_version=
   s.metadata = { "msys2_mingw_dependencies" => "sqlite3" } if s.respond_to? :metadata=
--- a/test/test_database.rb
+++ b/test/test_database.rb
@@ -217,6 +217,12 @@ module SQLite3
       assert db.closed?
     end
 
+    def test_close_v2
+      db = SQLite3::Database.new(':memory:')
+      db.close_v2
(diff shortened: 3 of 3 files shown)


flavorjones:
Thank you for submitting this PR! I've kicked off CI.

I'm curious if there's a reason we shouldn't just always call `close_v2`? Why even give users the option to not call it?

clemensp:
I initially wondered whether that was the desired move as well (i.e. `sqlite3_rb_close` calls `sqlite3_close_v2`).

My conclusion after dwelling on this:
`sqlite3_close_v2` is something you can call when you have certainty that you've called `finalize` on all statements, and you've finalized all backups (or haven't started any). This is because it assumes that everything will finalize eventually.

In my case, the app always calls the following:
https://github.com/sparklemotion/sqlite3-ruby/blob/69c9fa7f0013cfbc546f5b0256509156d0838f5d/lib/sqlite3/database.rb#L183

Which calls the block form of:
https://github.com/sparklemotion/sqlite3-ruby/blob/69c9fa7f0013cfbc546f5b0256509156d0838f5d/lib/sqlite3/database.rb#L151

Which ensures that the statement is closed:
https://github.com/sparklemotion/sqlite3-ruby/blob/69c9fa7f0013cfbc546f5b0256509156d0838f5d/lib/sqlite3/database.rb#L158

Which ensures that the statement is finalized:
https://github.com/sparklemotion/sqlite3-ruby/blob/69c9fa7f0013cfbc546f5b0256509156d0838f5d/ext/sqlite3/statement.c#L85


However, if someone chooses to use the non-block form of prepare:
https://github.com/sparklemotion/sqlite3-ruby/blob/69c9fa7f0013cfbc546f5b0256509156d0838f5d/lib/sqlite3/database.rb#L151

Then they would have to make sure they call close on the statement themselves.
So at the library level, we can't assume that all statements have been finalized. Otherwise I think we risk masking issues of unfinalized statements.

For that reason, I think we have to leave the decision to each client.

clemensp:
Also, this can be a separate issue/PR, but it might be worth exposing `sqlite3_next_stmt`:
https://sqlite.org/capi3ref.html#sqlite3_next_stmt

Discovered that in this thread:
https://sqlite-users.sqlite.narkive.com/zsHOze4n/sqlite-sqlite3-close-v2-leading-to-database-corruption

It would provide a tool to figure out what statements aren't finalized.


(BTW, I know the link's topic sounds alarming, but the issue itself looks to be usecase-specific. Re-using DB file names.)

clemensp:
@flavorjones : Does the change itself look okay?

I'm not super familiar with the Ruby <-> C interface, so the implementation of `sqlite3_rb_close_v2` is a copy of `sqlite3_rb_close`, but just calling `sqlite3_close_v2`. I figured the change was fine since as far as the Ruby side is concerned, the object can be de-allocated.

I'm asking because if it looks fine, then I'm looking to try a fix for the app I'm working on.
If there's anything else I can help with as well, please let me know. :)

flavorjones:
@clemensp OK, the code you're adding in this PR seems like it will work fine, so if you're blocked on me you should feel free to go ahead and try this out on your app.

I'll also note that this PR is based on `v1.4.4` and not `master`. You mentioned that you wanted to backport, and I'm curious why? I would prefer not to cut any more v1.4.x or v1.5.x releases.

Finally: I think there's a deeper change here that we should consider, which is whether `close_v2` should always be called (as well as adding some test coverage for it). I'd like to do a bit of exploratory work around the Ruby GC lifecycle and unfinalized database connections, etc. in general because it looks like they may be leaking memory.

clemensp:
@flavorjones : Thanks. I'll go ahead and test in my app.

re `v1.4.4`: My app currently is on that version, which is why my PR was based on it.
As for the backport, on second thought, it's not needed. I can point to a branch in my fork'd repo, and work on bumping up the gem version to be able to use the upstream gem.

As for always using `close_v2`, I left my comments here:
https://github.com/sparklemotion/sqlite3-ruby/pull/386#issuecomment-1517989294

flavorjones:
@clemensp OK, I think I have a grasp on what's probably going on.

In summary: we don't need (and shouldn't use) `close_v2`. I think it's likely the program you're working on is not closing all its resources. 

If you take a look at #387, I've added a new CI job that uses valgrind to check for memory leaks -- and I found a lot! But they were all due to unclosed `Database`, `Statement`, or `ResultSet` objects. I added `Database#close` calls and then the "unable to close due to unfinalized statements or unfinished backups" messages pointed me exactly to where I had unclosed statements or result sets. Once I properly closed all the objects, the memory leaks no longer existed and the valgrind job is clean.

If you don't mind, can you go back to the code and double-check that:

- all `Statement` objects returned from `Database#prepare` are closed properly
- all `ResultSet` objects returned from `Database#query` are closed properly

I think if you do you'll find you're able to call `Database#close` without an exception being raised.

clemensp:
Thanks @flavorjones . Will go back and look for unclosed objects.

Do you happen to have an example of how it shows where the unclosed statement is? I ask because the messages I'm logging atm don't point to the location. I presume it's because I'm not capturing enough in the logs.

flavorjones:
My advice would be to look for all instances of `Database#query` or `Database#prepare` and make sure they're being closed. Unfortunately it's not easy for sqlite-ruby to tell you where the query was made in the codebase.

PR #408, opened 2023-09-16
ext: compile sqlite with WAL_SYNCHRONOUS=1 and --enable-fts5

flavorjones:
See https://fractaledmind.github.io/2023/09/10/enhancing-rails-sqlite-optimizing-compilation/

Would love feedback @fractaledmind @hopsoft

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # sqlite3-ruby Changelog
 
+## next / unreleased
+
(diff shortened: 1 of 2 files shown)


oldmoe:
These look good, memstatus = 0 is usually another desirable setting (to reduce lock contention), but I guess since the gem doesn't release the GVL it wouldn't matter anyway

PR #415, opened 2023-10-14
make SQLite3::Database.open return the block result instead of the db instance

toy:
To make it follow common pattern, like for File:

    result = File.open('/dev/zero') do |f|
      f.read(10)
    end

See discussion in https://github.com/sparklemotion/sqlite3-ruby/discussions/414

### DIFF
--- a/test/test_database.rb
+++ b/test/test_database.rb
@@ -198,14 +198,36 @@ module SQLite3
 
     def test_new
       db = SQLite3::Database.new(':memory:')
-      assert db
+      assert_instance_of(SQLite3::Database, db)
     ensure
       db.close if db
+    end
+
+    def test_open
+      db = SQLite3::Database.open(':memory:')
+      assert_instance_of(SQLite3::Database, db)
+    ensure
+      db.close if db
+    end
+
+    def test_open_returns_block_result
+      result = SQLite3::Database.open(':memory:') do |db|
+        :foo
+      end
+      assert_equal :foo, result
     end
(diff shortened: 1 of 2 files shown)


test/test_database.rb
```diff
@@ -203,6 +203,20 @@ def test_new
       db.close if db
     end
 
+    def test_open
+      db = SQLite3::Database.open(':memory:')
+      assert db
```
flavorjones, line 208: Can you please assert on the type of this object?

flavorjones (changes requested):
I'm really happy with this PR. Thank you for the thoughtful additional test coverage. I have one small change I requested but otherwise I'd be happy to merge this.

test/test_database.rb
toy: I also changed the check for other places with just `assert db` and added one missing test - to ensure that [an instance of Database is yielded](https://github.com/sparklemotion/sqlite3-ruby/compare/adee4c7d2903e46318a4b7190590fdcd4ea50f28..0ef9a94d8f50122f59e45f2d7a535f02eefce5a3#diff-1a78f662d21da725643065914897a8dbd41fbd5893c3a90353382f367285330aR228-R235)

flavorjones: Thank you!

toy:
Interface tells me that I can merge, ok to do so?

flavorjones:
Thanks for your patience, merged!

PR #426, opened 2023-11-24
Support default transaction mode.

masamitsu-murase:
Let me propose a default transaction mode, which is used as a transaction mode when `mode` argument of `SQLite3::Database#transaction" is omitted.

This pull request does not break compatibility with the current version because the default value of `default_transaction_mode` is `:deferred`.

Usage:

```ruby
db = SQLite3::Database.new "database.db", default_transaction_mode: :immediate
db.transaction  # "begin immediate transaction" is issued.
db.commit
db.transaction :deferred  # "begin deferred transaction" is issued.
db.commit
```

This pull request also enables us to set the default transaction mode to `IMMEDIATE` or `EXCLUSIVE` in rails.

Regards,
Murase


flavorjones:
@masamitsu-murase Thank you for suggesting this feature! I'm not familiar with how the Rails adapter is evolving, but this seems like a reasonable change. I've kicked off CI.

masamitsu-murase:
Thank you so much!

PR #427, opened 2023-11-24
dep: update sqlite to v3.44.2

flavorjones:
https://sqlite.org/releaselog/3_44_2.html


flavorjones:
Truffleruby-head is failing for unrelated reasons, see https://github.com/oracle/truffleruby/issues/3328

PR #429, opened 2023-11-26
ext: add experimental support for ruby 3.3

flavorjones:


flavorjones:
Please note this won't get released until one of two things happens:

- the Ruby Core team freezes the ABI (binary interface) for Ruby 3.3 (see https://bugs.ruby-lang.org/issues/19980)
- Ruby 3.3.0 final release arrives

PR #431, opened 2023-11-27
dep: drop support for old rubies

flavorjones:
- drop support completely for Ruby `1.9.2 .. 2.6` (inclusive)
- drop native gem support for 2.7 (but keep general support for 2.7)

I'd really like feedback on dropping so many old versions. The honest truth is that we haven't tested anything older than 2.7 for the last year, anyway.


tenderlove (approved):
Seems fine to me!

PR #432, opened 2023-11-30
Convert SQLite3 objects to TypedData API

casperisfine:
The replacement API was introduced in Ruby 1.9.2 (2010),
and the old untyped data API was marked a deprecated in the documentation
as of Ruby 2.3.0 (2015)

Ref: https://bugs.ruby-lang.org/issues/19998

This also allows to implement `memsize` functions (even though sqlite opaque types makes this a bit pointless), and allow to enable immediate free and write barriers on most objects.

@flavorjones @tenderlove

### DIFF
--- a/ext/sqlite3/statement.c
+++ b/ext/sqlite3/statement.c
@@ -6,19 +6,29 @@
 
 VALUE cSqlite3Statement;
 
-static void deallocate(void * ctx)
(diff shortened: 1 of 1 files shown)


flavorjones (approved):
Thank you for doing this! It was on my TO-DO list :stuck_out_tongue_winking_eye:

Issue #434, opened 2023-12-12
"symbol not found" on alpine 3.19


rickselby:
**TL;DR from the maintainers -- read me first!**

You can work around this issue in two ways:

1. pin to alpine 3.18
2. compile the gem from source, see [INSTALLATION.md](https://github.com/sparklemotion/sqlite3-ruby/blob/master/INSTALLATION.md#avoiding-the-precompiled-native-gem)

Pick one! Permanent fix is being worked on upstream in rake-compiler-dock https://github.com/rake-compiler/rake-compiler-dock/issues/75

Thanks for your patience :heart: 

(The original bug description follows)

-----

On the latest ruby image with alpine 3.19, I can install `sqlite3` ruby gem, but it emits "symbol not found" errors at runtime.

This is a quick test on alpine 3.18:
```
$ docker run --rm -it ruby:3.2.2-alpine3.18 /bin/sh
/ # gem install sqlite3
Fetching sqlite3-1.6.9-x86_64-linux.gem
Successfully installed sqlite3-1.6.9-x86_64-linux
1 gem installed

A new release of RubyGems is available: 3.4.10 → 3.4.22!
Run `gem update --system 3.4.22` to update your installation.

/ # ruby -rsqlite3 -e 'puts SQLite3::SQLITE_LOADED_VERSION'
4.44.2
/ # 
```

And here's the same on 3.19:
```
$ docker run --rm -it ruby:3.2.2-alpine3.19 /bin/sh
/ # gem install sqlite3
Fetching sqlite3-1.6.9-x86_64-linux.gem
Successfully installed sqlite3-1.6.9-x86_64-linux
1 gem installed

A new release of RubyGems is available: 3.4.10 → 3.4.22!
Run `gem update --system 3.4.22` to update your installation.

/ # ruby -rsqlite3 -e 'puts SQLite3::SQLITE_LOADED_VERSION'
<internal:/usr/local/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:85:in `require': cannot load such file -- sqlite3/sqlite3_native (LoadError)
	from <internal:/usr/local/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
	from /usr/local/bundle/gems/sqlite3-1.6.9-x86_64-linux/lib/sqlite3.rb:6:in `rescue in <top (required)>'
[...]

PR #436, opened 2023-12-13
dep: rcd to 1.4.0.rc2 which uses ruby 3.3.0-rc1

flavorjones:


flavorjones:
musl failures are due to #434

Issue #288, opened 2020-05-10
'gem install sqlite3' fails on FreeBSD


caponecicero:
Gem installation fails on FreeBSD because the `Makefile` for the Ruby/C glue only looks  for `/usr/include/sqlite3.h` and `/usr/lib/libsqlite.so`.  On FreeBSD, the locations are `/usr/local/include/sqlite3.h` `/usr/local/lib/libsqlite3.so` and because **sqlite3** is installed through the `ports` system.

caponecicero:
[gem_make.out.txt](https://github.com/sparklemotion/sqlite3-ruby/files/4606424/gem_make.out.txt)

eddiec:
I install ruby into my home dir but this worked for me:
`gem  install sqlite3 -- --with-sqlite3-include=/usr/local/include/ --with-sqlite3-lib=/usr/local/lib`

flavorjones:
I would welcome documentation changes, or changes to the `extconf.rb` file to make the gem install more easily on FreeBSD. I would also welcome someone adding FreeBSD to the CI pipelines, maybe something similar to what Nokogiri does: https://github.com/sparklemotion/nokogiri/blob/main/.github/workflows/ci.yml#L334-L358

Worth noting, the `extconf.rb` has been completely rewritten since May 2020, and so we should make sure it's still a problem, and if

PR #437, opened 2023-12-13
ci: coverage for freebsd

flavorjones:
inspired by https://github.com/sparklemotion/sqlite3-ruby/issues/288


flavorjones:
This is linking against the system sqlite library when we try to build with the vendored library. I think this can be fixed by finished the work at https://github.com/sparklemotion/sqlite3-ruby/pull/405

PR #443, opened 2023-12-21
Add a busy_handler_timeout setter

fractaledmind:
One of the largest pain-points when using SQLite in a Rails app is the limited concurrency support. While SQLite does support only one writer, from a Rails app's point-of-view this does not mean that that app simply cannot run in, for example, Puma clustered mode. As I have [detailed](https://fractaledmind.github.io/2023/12/11/sqlite-on-rails-improving-concurrency/), the primary issues are that the GVL isn't released while the SQLite `busy_timeout` C function is running and transactions are run in `DEFERRED` mode. 

We can solve the first of these by providing a Ruby `busy_handler` callback that will release the GVL between retries. This allows multiple threads to naturally coordinate and "queue up" to acquire SQLite's single write lock.

I initially proposed providing this in the Rails codebase (see: https://github.com/rails/rails/pull/50370); however, it was suggested that this functionality more naturally belongs in this codebase. So, I would love to start brainstorming the interface for this. Once built and released, we can then update Rails to use this instead of the `busy_timeout`. 

Jean Boussier suggested `lock_wait_timeout` as a name. I'm opening this PR to get the ball rolling. I'm open to naming suggestions as well as test suggestions.

### DIFF
--- a/lib/sqlite3/database.rb
+++ b/lib/sqlite3/database.rb
@@ -131,6 +131,7 @@ module SQLite3
       @type_translator  = make_type_translator @type_translation
       @readonly         = mode & Constants::Open::READONLY != 0
       @default_transaction_mode = options[:default_transaction_mode] || :deferred
+      @timeout_deadline = nil
 
       if block_given?
         begin
@@ -697,12 +698,16 @@ module SQLite3
     # while SQLite sleeps and retries.
     def busy_handler_timeout=( milliseconds )
       timeout_seconds = milliseconds.fdiv(1000)
-      timeout_deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + timeout_seconds
 
       busy_handler do |count|
-        next false if Process.clock_gettime(Process::CLOCK_MONOTONIC) > timeout_deadline
-
-        sleep(0.001)
+        now = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+        if count.zero?
+          @timeout_deadline = now + timeout_seconds
+        elsif now > @timeout_deadline
+          next false
+        else
+          sleep(0.001)
+        end
       end
     end
 
--- a/lib/sqlite3/database.rb
+++ b/lib/sqlite3/database.rb
@@ -691,7 +691,7 @@ module SQLite3
       @readonly
     end
 
-    def lock_wait_timeout=( milliseconds )
+    def busy_handler_timeout=( milliseconds )
       timeout_seconds = milliseconds.fdiv(1000)
       retry_interval = 0.001 # 1 millisecond
 
--- a/lib/sqlite3/database.rb
+++ b/lib/sqlite3/database.rb
@@ -691,16 +691,19 @@ module SQLite3
       @readonly
     end
 
+    # Sets a #busy_handler that releases the GVL between retries,
+    # but only retries up to the indicated number of +milliseconds+.
+    # This is an alternative to #busy_timeout, which holds the GVL
+    # while SQLite sleeps and retries.
     def busy_handler_timeout=( milliseconds )
       timeout_seconds = milliseconds.fdiv(1000)
       retry_interval = 0.001 # 1 millisecond
+      timeout_deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + timeout_seconds
 
       busy_handler do |count|
-        @start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC) if count == 0
+        next false if Process.clock_gettime(Process::CLOCK_MONOTONIC) > timeout_deadline
 
-        next sleep(retry_interval) unless (count % 100) == 0
-
-        (Process.clock_gettime(Process::CLOCK_MONOTONIC) - @start_time) <= timeout_seconds
+        sleep(retry_interval)
       end
     end
 
--- a/lib/sqlite3/database.rb
+++ b/lib/sqlite3/database.rb
@@ -697,13 +697,12 @@ module SQLite3
     # while SQLite sleeps and retries.
     def busy_handler_timeout=( milliseconds )
       timeout_seconds = milliseconds.fdiv(1000)
-      retry_interval = 0.001 # 1 millisecond
       timeout_deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + timeout_seconds
 
       busy_handler do |count|
         next false if Process.clock_gettime(Process::CLOCK_MONOTONIC) > timeout_deadline
 
-        sleep(retry_interval)
+        sleep(0.001)
       end
     end
 
--- a/lib/sqlite3/database.rb
+++ b/lib/sqlite3/database.rb
@@ -691,6 +691,19 @@ module SQLite3
       @readonly
     end
 
+    def lock_wait_timeout=( milliseconds )
+      timeout_seconds = milliseconds.fdiv(1000)
+      retry_interval = 0.001 # 1 millisecond
+
+      busy_handler do |count|
+        @start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC) if count == 0
+
+        next sleep(retry_interval) unless (count % 100) == 0
+
+        (Process.clock_gettime(Process::CLOCK_MONOTONIC) - @start_time) <= timeout_seconds
+      end
+    end
+
     # A helper class for dealing with custom functions (see #create_function,
     # #create_aggregate, and #create_aggregate_handler). It encapsulates the
     # opaque function object that represents the current invocation. It also
--- a/test/test_integration_pending.rb
+++ b/test/test_integration_pending.rb
@@ -114,64 +114,36 @@ class TC_Integration_Pending < SQLite3::TestCase
   end
 
   def test_busy_timeout_blocks_gvl
-    t1 = Thread.new do
-      begin
-        db1 = SQLite3::Database.open( "test.db" )
-        db1.transaction( :exclusive ) do
-          sleep 0.05
-        end
-      ensure
-        db1.close if db1
-      end
-    end
-
-    t2 = Thread.new do
-      begin
-        db2 = SQLite3::Database.open( "test.db" )
-        db2.busy_timeout 1000
-        db2.transaction( :exclusive ) do
-          sleep 0.01
-        end
-      ensure
-        db2.close if db2
-      end
-    end
+    db2 = SQLite3::Database.open( "test.db" )
+    db2.busy_timeout 1000
 
     time = Benchmark.measure do
-      [t1, t2].each(&:join)
+      @db.transaction( :exclusive ) do
+        assert_raises( SQLite3::BusyException ) do
+          db2.transaction( :exclusive ) {}
+        end
+      end
     end
 
-    assert_operator time.real, :>=, 1
+    db2.close
+
+    assert_operator time.real, :>=, 2
   end
 
   def test_busy_handler_timeout_releases_gvl
-    t1 = Thread.new do
-      begin
-        db1 = SQLite3::Database.open( "test.db" )
-        db1.transaction( :exclusive ) do
-          sleep 0.05
-        end
-      ensure
-        db1.close if db1
-      end
-    end
-
-    t2 = Thread.new do
-      begin
-        db2 = SQLite3::Database.open( "test.db" )
-        db2.busy_handler_timeout = 1000
-        db2.transaction( :exclusive ) do
-          sleep 0.01
-        end
-      ensure
-        db2.close if db2
-      end
-    end
+    db2 = SQLite3::Database.open( "test.db" )
+    db2.busy_handler_timeout = 1000
 
     time = Benchmark.measure do
-      [t1, t2].each(&:join)
+      @db.transaction( :exclusive ) do
+        assert_raises( SQLite3::BusyException ) do
+          db2.transaction( :exclusive ) {}
+        end
+      end
     end
 
-    assert_operator time.real, :<, 1
+    db2.close
+
+    assert_operator time.real, :>=, 1
   end
 end
--- a/test/test_integration_pending.rb
+++ b/test/test_integration_pending.rb
@@ -129,9 +129,6 @@ class TC_Integration_Pending < SQLite3::TestCase
       begin
         db2 = SQLite3::Database.open( "test.db" )
         db2.busy_timeout 1000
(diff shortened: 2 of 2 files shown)


fractaledmind:
@suwyn Do you happen to have a better alternative name for this? I know that you have been thinking about this problem space for a while, so I imagine you might have some language around distinguishing a Ruby `busy_timeout` from the SQLite/C `busy_timeout` in a way that is succinct but communicative.

suwyn:
Oh boy, naming is hard!  Maybe `busy_handler_timeout`?

fractaledmind:
I like that. I'll switch to that

suwyn:
@fractaledmind Also consider we'll want to configure this from the database config file so adding it as an option would be beneficial.  Here is [an example](https://github.com/sparklemotion/sqlite3-ruby/pull/426/files#diff-a5f0046d65327d07a9b89f22eae22a1d9e609e03a3e17f82ada91ec6e0d1943aR133) from #426

Default of `nil` which wouldn't set the busy handler.

suwyn:
Also noting for prosperity why this is needed.   The underlining issue with the interpreter is [documented here](http://nerdjusttyped.blogspot.com/2014/11/threaded-sqlite-access-and-ruby-sqlite.html).

From my experiments there is a lot of speed to be gained if there was a way to fix the underlining issue.

fractaledmind:
This will be set in Rails the same way that `busy_timeout` is currently set. Rails' `database.yml` won't change at all, it will just be that the `timeout` value will be passed to this function and not the `busy_timeout` function.

flavorjones:
Ignore any CI failures for "native packaging" with "head", that's only because the HEAD version rolled over to 3.4.dev.

lib/sqlite3/database.rb
```diff
@@ -691,6 +691,19 @@ def readonly?
...
     end
 
+    def busy_handler_timeout=( milliseconds )
+      timeout_seconds = milliseconds.fdiv(1000)
+      retry_interval = 0.001 # 1 millisecond
+
+      busy_handler do |count|
+        @start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC) if count == 0
+
+        next sleep(retry_interval) unless (count % 100) == 0
```
byroot, line 701: I'm not even sure this module thing is worth it. `Process.clock_gettime(Process::CLOCK_MONOTONIC)` is really really fast, it's not even really a syscall as it uses VDSO so never leaves user space (on Linux).

I just benchmarked it, it's off course slower than doing a modulo, but it's in the same order of magnitude: https://gist.github.com/casperisfine/72af4812a0cc8c925432a5d2722bc12a

So given that every sleep could be very long, I don't think this potential for innacuracy is worth this micro-optimization. You can just pre-compute the deadline like in my benchmark, and that should be more than fast enough.

Also if we're going for micro-optimizations, might as well define that instance variable early to ensure a consistent shape.

fractaledmind: @byroot: Makes sense. The win from releasing the GVL is the core detail, and any other micro-optimizations should be considered later based on real-world usage. I simplified the handler in [d3dcd52](https://github.com/sparklemotion/sqlite3-ruby/pull/443/commits/d3dcd52c753462653d2584f93fea0db43018d172)

lib/sqlite3/database.rb
```diff
@@ -691,6 +691,22 @@ def readonly?
       @readonly
     end
 
+    # Sets a #busy_handler that releases the GVL between retries,
+    # but only retries up to the indicated number of +milliseconds+.
+    # This is an alternative to #busy_timeout, which holds the GVL
+    # while SQLite sleeps and retries.
+    def busy_handler_timeout=( milliseconds )
+      timeout_seconds = milliseconds.fdiv(1000)
+      retry_interval = 0.001 # 1 millisecond
```
byroot, line 700: You can do without that local variable, that will be one less indirection in the proc.

fractaledmind:
@djmb @rosa: I was just reading thru the [SolidQueue](https://github.com/basecamp/solid_queue/) source as I work on integrating it into a project I have at work. I came across [this commit](https://github.com/basecamp/solid_queue/commit/08f5725a5aa3f5169305baafb88db4a592df5086) that patches Rails' SQLite3 adapter to improve concurrency. Reading the commit description, it sounds like you have had similar experiences as I have working to get SQLite to handle concurrency in a reasonable and resilient way. 

I note that in your comment and patch, you lean on the `retries` option and patch it to have linear backoff (`sleep`ing `count` milliseconds on each `busy_handler` invocation). I'd love to hear more about what, in particular, was causing problems with the current Rails implementation of `retries` that doesn't backoff between retries. I'd also be keen to get your perspective on this implementation of a `busy_handler` and the general plan for Rails. To recap conversations had across repos, the plan as it stands currently is:

1. get this PR merged into the SQLite3 driver to provide a `busy_handler_timeout` method that accepts a timeout in milliseconds, but uses a Ruby `busy_handler` instead of the SQLite C `busy_timeout` to ensure that the GVL is released between retries
2. deprecate the `retries` option in Rails for the SQLite adapter
3. have the `timeout` option in Rails for the SQLite adapter call this `busy_handler_timeout` method from the SQLite3 driver instead of the `busy_timeout` method

After working with the `retries` option for some time, and getting feedback from the community, I believe that the mental model of `timeout` is clearer and people better understand what kind of value to set for their needs. However, we need to deal with the GVL lock issue of the `busy_timeout` method in this driver. Does the current plan make sense given you and your team's experience with SQLite usage in real-world apps? Does the current implementation of the `busy_handler_timeout` here make sense given you and your team's experience?

I believe your perspective could help ensure that both this PR and the larger plan for Rails' SQLite support are as solid as possible.

fractaledmind:
@flavorjones @byroot: Added two simple tests focusing on the GVL blocking difference between `busy_timeout` and `busy_handler_timeout`. AppVeyor is green now. Is there anything still blocking here?

lib/sqlite3/database.rb
```diff
@@ -691,6 +691,21 @@ def readonly?
       @readonly
     end
 
+    # Sets a #busy_handler that releases the GVL between retries,
+    # but only retries up to the indicated number of +milliseconds+.
+    # This is an alternative to #busy_timeout, which holds the GVL
+    # while SQLite sleeps and retries.
+    def busy_handler_timeout=( milliseconds )
+      timeout_seconds = milliseconds.fdiv(1000)
+      timeout_deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + timeout_seconds
```
suwyn, line 700: @fractaledmind Looking at the code again, I believe there is a bug here.   

This is setting a clock based timeout deadline at the time `busy_handler_timeout` is set rather than when the busy_handler gets invoked for the first time.   Essentially setting a hard timeout for `timeout_seconds` for all invocations of the busy handler.

Consider the following (I didn't test it but I believe this should show the issue, if I am reading the code correctly)

```
busy_handler_timeout_db = SQLite3::Database.open( "test.db" )
busy_handler_timeout_db.busy_handler_timeout = 1000
sleep 1.5

# timeout_deadline should be expired now since that was set 
# at the point in time when we set busy_handler_timeout to 1000.
# Any invocations of the busy handler will now throw busy errors
```

fractaledmind: I think you're right. Great catch. Will fix when I'm back at my laptop

fractaledmind: @suwyn Checkout https://github.com/sparklemotion/sqlite3-ruby/pull/443/commits/48daae13fb32213c84a07ea7e7ed8714d0462fb6

suwyn: Looks better @fractaledmind 

It still has me wondering if it's thread safe though since it is using an instance variable.   If two threads used the same connection, you'd have an issue.   I did some experimenting in Rails and the connection pool appears to be thread safe (as it states in the docs) .

Experimenting with the gem on its own, I wasn't able quickly able to cause any issue.   SQLite doesnt have a way to sleep so I stopped short of writing a long running UDF.  

I think the ideal solution is to either bake the busy handler into the C code or pass the time of first invocation into the handler from C, like it does with count.   I'd obviously strongly prefer the former, but have no idea on what that would take and outside of the scope of this PR.

So short of using the count to estimate the clock time (e.g. the `(count * RETRY_INTERVAL) > timeout_seconds` approach) instead of the clock time we'll just have to know that a race condition probably exists.

fractaledmind: @suwyn: You should join this Discord server where we talk a lot about the SQLite ecosystem in Ruby: https://discord.gg/ehdDh5C4

That would allow us to connect more easily and even pair on this. I'd love some help on writing useful tests. I have a long running UDF already written:
```sql
WITH RECURSIVE r(i) AS (
  VALUES(0)
  UNION ALL
  SELECT i FROM r
  LIMIT 10000000
)
SELECT i FROM r ORDER BY i LIMIT 1;
```
What I don't have are resilient and expansive tests. Want to work on getting this over the line together?

djmb:
> I note that in your comment and patch, you lean on the retries option and patch it to have linear backoff (sleeping count milliseconds on each busy_handler invocation). I'd love to hear more about what, in particular, was causing problems with the current Rails implementation of retries that doesn't backoff between retries. 

Hi @fractaledmind! The issue with using retries without a backoff was that we would still get `SQLite3::BusyException`s.

I've just tried it out with no sleep and to avoid the BusyExceptions, I need to set the retries to about 1,000,000, at which point it seems to hang. I don't know exactly what the mechanism is there though - whether the thread holding the lock is actually blocked or if it is being starved.

> I'd also be keen to get your perspective on this implementation of a busy_handler and the general plan for Rails. To recap conversations had across repos, the plan as it stands currently is:
> 
> get this PR merged into the SQLite3 driver to provide a busy_handler_timeout method that accepts a timeout in milliseconds, but uses a Ruby busy_handler instead of the SQLite C busy_timeout to ensure that the GVL is released between retries
> deprecate the retries option in Rails for the SQLite adapter
> have the timeout option in Rails for the SQLite adapter call this busy_handler_timeout method from the SQLite3 driver instead of the busy_timeout method
> After working with the retries option for some time, and getting feedback from the community, I believe that the mental model of timeout is clearer and people better understand what kind of value to set for their needs. However, we need to deal with the GVL lock issue of the busy_timeout method in this driver. Does the current plan make sense given you and your team's experience with SQLite usage in real-world apps? Does the current implementation of the busy_handler_timeout here make sense given you and your team's experience?
> 
> I believe your perspective could help ensure that both this PR and the larger plan for Rails' SQLite support are as solid as possible.

We only used `retries` here because the current `timeout` wouldn't release the GVL lock, so your plan here sounds perfect. I agree that `timeout` is much more intuitive. 

It seems likely we will still need a sleep in the handler to avoid the issues we've been seeing, but it would be good to know why. I'll see if I can debug what's going on in Solid Queue myself. It's tricky because the I/O of logging anything from the handler seems to have the same effect as a sleep and the BusyExceptions go away.

If you want to dig about yourself, the patch for sleeping is to allow the tests to pass, you can run them with `TARGET_DB=sqlite rails test`. Removing the patch should give you the BusyExceptions.

lib/sqlite3/database.rb
```diff
@@ -691,6 +692,25 @@ def readonly?
...
+      timeout_seconds = milliseconds.fdiv(1000)
+
+      busy_handler do |count|
+        now = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+        if count.zero?
+          @timeout_deadline = now + timeout_seconds
+        elsif now > @timeout_deadline
+          next false
+        else
+          sleep(0.001)
```
tenderlove, line 709: Is this sleep necessary? I think the VM is interruptible on any method call so the calls to clock_gettime etc should do the trick (I think).

tenderlove (approved):
I'm fine with the patch if @byroot is OK with it. I had one question wrt the `sleep` (but including the sleep is fine if it's necessary)

lib/sqlite3/database.rb
byroot: @tenderlove it's not about being interruptible, it's about forcibly releasing the GVL. The assumption is that the client that is currently holding the SQLite3 write lock might be another thread in the same process, hence we should try to switch back to it rather than busy loop for 100ms until the thread scheduler quantum is reached.

And even if it's not in that process, yielding the GVL allow other unrelated threads to proceed.

NB: for the "in same process case" a shared Ruby mutex would be way more efficient, but we'd need some way to tell two clients are pointing that the same database, hence should share a single Mutex.

tenderlove: > The assumption is that the client that is currently holding the SQLite3 write lock might be another thread in the same process, hence we should try to switch back to it

I see. In that case wouldn't `Thread.pass` also do the trick?

byroot: It would yes, but then the risk is that there's no other ready thread, causing the process to essentially be busy looping, which would pin one CPU to 100% and may not be desirable. A very short sleep actually make sense IMO.

tenderlove: Sounds good to me!

flavorjones:
@fractaledmind this is causing failures in CI on windows

```
  1) Failure:
TC_Integration_Pending#test_busy_timeout_blocks_gvl [D:/a/sqlite3-ruby/sqlite3-ruby/test/test_integration_pending.rb:100]:
SQLite3::BusyException expected but nothing was raised.
```

see e.g. https://github.com/sparklemotion/sqlite3-ruby/actions/runs/7403622873/job/20143757496

Can you please take a look? Or let me know if you'd prefer me to revert.

fractaledmind:
I have been trying to debug, but can't reproduce. I am now trying to research a more direct way to test "holds GVL" vs "releases GVL" for these two. You can revert for now, and I will try to rework the tests. But I can only really see if new tests are good by running thru CI as locally things are ok. 

On a related note, is there a way to ensure local tests are running with current head version of SQLite? It seems my tests run against my default macOS version of SQLite.

flavorjones:
> On a related note, is there a way to ensure local tests are running with current head version of SQLite? It seems my tests run against my default macOS version of SQLite.

By default this should not happen. Try:

```sh
bundle exec rake clean clobber
bundle exec rake compile
bundle exec rake test
```

If you're still using the system libraries, open a new issue and attach

- full output from these commands
- the Makefile and mkmf.log from the compilation phase (which should be under `./tmp/x86_64-darwin/sqlite3_native/3.2.2` or similarly-named directory)

PR #452, opened 2024-01-03
style(c): adopt cruby conventions

flavorjones:
and some other conventions that we're using in Nokogiri.

### DIFF
--- a/ext/sqlite3/sqlite3.c
+++ b/ext/sqlite3/sqlite3.c
@@ -3,162 +3,167 @@
 VALUE mSqlite3;
 VALUE cSqlite3Blob;
 
-int bignum_to_int64(VALUE value, sqlite3_int64 *result)
+int
+bignum_to_int64(VALUE value, sqlite3_int64 *result)
 {
 #ifdef HAVE_RB_INTEGER_PACK
-  const int nails = 0;
-  int t = rb_integer_pack(value, result, 1, sizeof(*result), nails,
-			  INTEGER_PACK_NATIVE_BYTE_ORDER|
-			  INTEGER_PACK_2COMP);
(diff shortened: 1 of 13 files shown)


ext/sqlite3/sqlite3.c
```diff
@@ -3,162 +3,167 @@
...
-  case +1:
-    if (!nails) {
-      if (*result < 0) return 0;
+    const int nails = 0;
+    int t = rb_integer_pack(value, result, 1, sizeof(*result), nails,
+                            INTEGER_PACK_NATIVE_BYTE_ORDER |
+                            INTEGER_PACK_2COMP);
+    switch (t) {
+        case -2:
+        case +2:
```
tenderlove, line 16: `case` should be indented 2, and then 2 below the switch [like this](https://github.com/ruby/ruby/blob/master/prism_compile.c#L335-L349).

Vim configuration is this: `cinoptions=:2,=2,l1`

tenderlove (commented):
Indentation on CRuby is a little different for switch statements. I'm not married to that style though. What are your thoughts?

ext/sqlite3/sqlite3.c
flavorjones: I'd be fine with this if I knew how to make astyle do it (I don't think it can) or how to make .editorconfig describe it (I don't think it can).

:shrug:

flavorjones: yeah even clang-format doesn't support this indentation style

https://clang.llvm.org/docs/ClangFormatStyleOptions.html#indentcaselabels

If that's the only objection to this formatting I'd prefer to ship it and we can tweak it later.

Issue #287, opened 2020-04-03
exec_batch does not release GVL.


ioquatix:
```
#!/usr/bin/env ruby

require 'sqlite3'

Thread.new do
	while true
		sleep 0.1
		$stdout.write '.'
	end
end

s = "SELECT 1;"*500000
db = SQLite3::Database.new(':memory:')

sleep 1

$stdout.write '>'
db.execute_batch2(s)
$stdout.write '<'

sleep 1
```

Gives

```
koyoko% ./sqlite3.rb
..........>.<..........
```

Expected:

```
koyoko% ./sqlite3.rb
..........>............<..........
```

Or something to that effect.

cc @ko1 @tenderlove

tenderlove:
`execute_batch2` [calls back in to Ruby for each row](https://github.com/sparklemotion/sqlite3-ruby/blob/b98034d1fdaa94cd4ae76bbfa0ac56dd8dcc3cf5/ext/sqlite3/database.c#L731-L733).

We would have to unlock the GVL before calling `sqlite3_exec`, then re-acquire the lock each time we process a row.  AFAIK releasing / acquiring the lock isn't free, so this could slow down overall throughput (especially if done for every row).  On top of that, I don't think anyone actually uses this API, let alone in a multi-threaded application.  Another challenge is that SQLite3 is only [optionally thread safe](https://www.sqlite.org/threadsafe.html), so we would have to add conditionals around whether or not the release the GVL (and those conditionals would have to exactly match SQLite3 which lets you pick thread safetyness at compile and/or runtime).  Finally, if someone were actually using this method in a threaded environment, and we made a mistake with the GVL lock / unlock we could corrupt their data.

For the above reasons, (high risk, low reward) I'm not particularly motivated to fix this.  So if someone wants to try tackling it, that would be great.  Otherwise I'm going to wait for user demand.

ioquatix:
I think what you proposed is the correct approach as locking the VM is definitely the wrong approach. The overhead of the per-row lock is to be worn by the user of the interface and is a natural consequence of the GVL.

Sqlite3 should not be used on multiple threads IMHO, it should be a separate connection/instance per thread. Does that work?

tenderlove:
> I think what you proposed is the correct approach as locking the VM is definitely the wrong approach.

I don't understand? My proposal is to do nothing unless there is demand for the feature by people that use it.

tenderlove:
[...]

PR #458, opened 2024-01-07
Implement progress handler interface

fractaledmind:
This adds support for the [`progress_handler`](https://www.sqlite.org/c3ref/progress_handler.html) SQLite interface. Such a callback is useful as it allows for statement timeouts to be defined. For example, we could define a `statement_timeout=` method on the database instance like so:
```ruby
def statement_timeout=( milliseconds )
  timeout_seconds = milliseconds.fdiv(1000)

  progress_handler do
    now = Process.clock_gettime(Process::CLOCK_MONOTONIC)
    if @statement_timeout_deadline.nil?
      @statement_timeout_deadline = now + timeout_seconds
    elsif now > @statement_timeout_deadline
      next false
    else
      true
    end
  end
end
```

Note: I don't know C, so this PR was written using pattern matching (primarily on the `rb_sqlite3_busy_handler` and `busy_handler` methods), some Googling (for how to get the correct format for `rb_scan_args`), and ChatGPT (to explain what all of this C code was doing).

### DIFF
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -237,6 +237,51 @@ static VALUE busy_handler(int argc, VALUE *argv, VALUE self)
   return self;
 }
 
+static int
+rb_sqlite3_progress_handler(void *ctx)
+{
+  VALUE self = (VALUE)(ctx);
+  VALUE handle = rb_iv_get(self, "@progress_handler");
+  VALUE result = rb_funcall(handle, rb_intern("call"), 0);
+
+  if (Qfalse == result) return 1;
+
+  return 0;
+}
+
+/* call-seq:
+ *    progress_handler([n]) { ... }
+ *    progress_handler([n,] Class.new { def call; end }.new)
+ *
+ * Register a progress handler with this database instance.
+ * This handler will be invoked periodically during a long-running query or operation.
+ * If the handler returns +false+, the operation will be interrupted; otherwise, it continues.
+ * The parameter 'n' specifies the number of SQLite virtual machine instructions between invocations.
+ * If 'n' is not provided, the default value is 1.
+ */
+static VALUE
+progress_handler(int argc, VALUE *argv, VALUE self)
+{
+  sqlite3RubyPtr ctx;
+  VALUE block, n_value;
+
+  TypedData_Get_Struct(self, sqlite3Ruby, &database_type, ctx);
+  REQUIRE_OPEN_DB(ctx);
+
+  rb_scan_args(argc, argv, "02", &n_value, &block);
+
+  int n = NIL_P(n_value) ? 1 : NUM2INT(n_value);
+  if(NIL_P(block) && rb_block_given_p()) block = rb_block_proc();
+
+  rb_iv_set(self, "@progress_handler", block);
+
+  sqlite3_progress_handler(
+      ctx->db, n, NIL_P(block) ? NULL : rb_sqlite3_progress_handler, (void *)self);
+
+  return self;
+}
+
+
 /* call-seq: last_insert_row_id
  *
  * Obtains the unique row ID of the last row to be inserted by this Database
@@ -854,6 +899,7 @@ void init_sqlite3_database(void)
   rb_define_method(cSqlite3Database, "changes", changes, 0);
   rb_define_method(cSqlite3Database, "authorizer=", set_authorizer, 1);
   rb_define_method(cSqlite3Database, "busy_handler", busy_handler, -1);
+  rb_define_method(cSqlite3Database, "progress_handler", progress_handler, -1);
   rb_define_method(cSqlite3Database, "busy_timeout=", set_busy_timeout, 1);
   rb_define_method(cSqlite3Database, "extended_result_codes=", set_extended_result_codes, 1);
   rb_define_method(cSqlite3Database, "transaction_active?", transaction_active_p, 0);
--- a/lib/sqlite3/database.rb
+++ b/lib/sqlite3/database.rb
@@ -124,6 +124,7 @@ module SQLite3
       @authorizer       = nil
       @encoding         = nil
       @busy_handler     = nil
+      @progress_handler = nil
       @collations       = {}
       @functions        = {}
       @results_as_hash  = options[:results_as_hash]
--- a/test/test_integration.rb
+++ b/test/test_integration.rb
@@ -470,7 +470,7 @@ class TC_Database_Integration < SQLite3::TestCase
 
   def test_transaction_implicit_rollback
     assert !@db.transaction_active?
-    @db.transaction 
+    @db.transaction
     @db.execute('create table bar (x CHECK(1 = 0))')
     assert @db.transaction_active?
     assert_raises( SQLite3::ConstraintException ) do
@@ -503,5 +503,57 @@ class TC_Database_Integration < SQLite3::TestCase
     result = @db.get_first_value( "select b from foo where a=? and b=?",
       [ 1, "foo" ] )
     assert_equal "foo", result
+  end
+
+  def test_progress_handler_used
+    progress_calls = []
+    @db.progress_handler do
+      progress_calls << nil
+      true
+    end
+    @db.execute "create table test1(a, b)"
+
+    assert_operator 1, :<, progress_calls.size
+  end
+
+  def test_progress_handler_opcode_arg
+    progress_calls = []
+    handler = Proc.new do
+      progress_calls << nil
+      true
+    end
+    @db.progress_handler(1, handler)
+    @db.execute "create table test1(a, b)"
+    first_count = progress_calls.size
+
+    progress_calls = []
+    @db.progress_handler(2, handler)
+    @db.execute "create table test2(a, b)"
+    second_count = progress_calls.size
+
+    assert_operator first_count, :>, second_count
+  end
+
+  def test_progress_handler_interrupts_operation
+    @db.progress_handler do
+      false
+    end
+
+    assert_raises(SQLite3::InterruptException) do
+      @db.execute "create table test1(a, b)"
+    end
+  end
+
+  def test_clear_handler
+    progress_calls = []
+    @db.progress_handler do
+      progress_calls << nil
+      true
+    end
+    @db.progress_handler(nil)
+
+    @db.execute "create table test1(a, b)"
+
+    assert_equal 0, progress_calls.size
   end
 end


fractaledmind:
[`cruby-x86_64-darwin-install (3.3)`](https://github.com/sparklemotion/sqlite3-ruby/pull/458#logs) timed out it seems. Can we rerun?

oldmoe:
+1 to adding the progress handler, maybe we need to make sure this doesn't explode if sqlite is omitting the progress handler in the compilation process

fractaledmind:
> +1 to adding the progress handler, maybe we need to make sure this doesn't explode if sqlite is omitting the progress handler in the compilation process

@oldmoe: Does c0d5144 resolve this issue do you think?

ext/sqlite3/database.c
```diff
@@ -251,6 +251,51 @@ busy_handler(int argc, VALUE *argv, VALUE self)
...
+
+  rb_scan_args(argc, argv, "02", &n_value, &block);
+
+  int n = NIL_P(n_value) ? 1 : NUM2INT(n_value);
+  if(NIL_P(block) && rb_block_given_p()) block = rb_block_proc();
+
+  rb_iv_set(self, "@progress_handler", block);
+
+  sqlite3_progress_handler(
+      ctx->db, n, NIL_P(block) ? NULL : rb_sqlite3_progress_handler, (void *)self);
```
byroot, line 293: You can't store a `VALUE` in C-land like this without it being marked, as the object might be moved by the GC, and end up pointing to either garbage or another object.

You can likely cause a crash by calling [`GC.verify_compaction_references`](https://github.com/msgpack/msgpack-ruby/blob/e52d18bacc5a69bd0c823e8daf2792b187a6be46/spec/spec_helper.rb#L21-L25) from inside the callback.

Instead you'd have to store that reference inside the `sqlite3Ruby` struct, and have it marked with `gc_mark` so it's pinned.

fractaledmind: I _think_ I understand what I should do, but to confirm: would this not also be a problem for the `busy_handler` implementation, from which this line of code is nearly identically copied (I just added the `n` argument). I do not see that object stored inside the `sqlite3Ruby` struct. If the `busy_handler` implementation is safe, could you explain to me the mechanisms that make it safe so that I can mirror those mechanisms for `progress_handler`?

byroot: No you are right, the busy handler has the same issue, which I hadn't noticed, it also need to be fixed. To be fair however the busy handler was added 15 years ago, when compaction wasn't a concern.

fractaledmind: Not a judgement of the code, just looking to find pre-existing solutions to mirror if possible. But it sounds like this PR will need to be the first implementation of compaction-resilient handling of such code, which we can then bring back to the `busy_handler`. I'll work on an initial attempt.

fractaledmind: This is definitely stretching me beyond my current very basic understanding of C and writing C extensions for Ruby.

I just pushed 35a003a which doesn't compile. On my Mac, I get this error and warning:
```
../../../../ext/sqlite3/database.c:49:5: warning: variable 'ctx' is uninitialized when used here [-Wuninitialized]
    ctx->progress_handler = Qnil;
    ^~~
../../../../ext/sqlite3/database.c:48:23: note: initialize the variable 'ctx' to silence this warning
    sqlite3RubyPtr ctx;
                      ^
                       = NULL
../../../../ext/sqlite3/database.c:262:32: error: member reference base type 'void' is not a structure or union
  VALUE result = rb_funcall(ctx->progress_handler, rb_intern("call"), 0);
                            ~~~^ ~~~~~~~~~~~~~~~~
1 warning and 1 error generated.
```

I also don't know how or where to call `rb_gc_mark` with the value object. My hope is that you can help me polish this initial, broken C code @byroot.

byroot: I can open a PR to fix the busy handler later today if you want.

byroot: Here you go: https://github.com/sparklemotion/sqlite3-ruby/pull/466

To understand that PR I recommend Peter excellent blog series: https://blog.peterzhu.ca/ruby-c-ext-part-7/

byroot:
You should start by rebasing on my PR, you'd have the new functions lined up etc.

fractaledmind:
I had a stray line in the `allocate` method. Removing that brings me to the same problems with TruffleRuby. Whenever you figure out how to solve that in your branch, I will rebase and bring that fix in then.

byroot:
> Removing that brings me to the same problems with TruffleRuby. Whenever you figure out how to solve that in your branch

Ah, I didn't notice there was an issue with TruffleRuby, seems pretty easy to fix though:

> xternal LLVMFunction rb_gc_mark_movable cannot be found.

Need to test for that function. I'll also report it to Truffle, as it should be trivial for them to add it.

ext/sqlite3/database.c
fractaledmind: Mirrored solution in https://github.com/sparklemotion/sqlite3-ruby/pull/466

fractaledmind:
All green. Thanks so much for the guidance @byroot.

fractaledmind:
Never seen this error before, but it looks to be completely unrelated to this PR (from `sqlcipher (windows-latest, mingw)`):
```
NOTE: ruby_memcheck is not available in this environment: cannot load such file -- ruby_memcheck
D:/a/sqlite3-ruby/sqlite3-ruby/vendor/bundle/ruby/3.4.0+0/gems/minitest-5.20.0/lib/minitest.rb:3: warning: mutex_m was loaded from the standard library, but is not part of the default gems since Ruby 3.4.0. Add mutex_m to your Gemfile or gemspec.
D:/ruby-mingw/lib/ruby/3.4.0+0/bundled_gems.rb:74:in `require': cannot load such file -- mutex_m (LoadError)
```

ext/sqlite3/database.c
```diff
@@ -259,6 +261,57 @@ busy_handler(int argc, VALUE *argv, VALUE self)
...
+{
+  sqlite3RubyPtr ctx;
+  VALUE block, n_value;
+
+  TypedData_Get_Struct(self, sqlite3Ruby, &database_type, ctx);
+  REQUIRE_OPEN_DB(ctx);
+
+  rb_scan_args(argc, argv, "02", &n_value, &block);
+
+  int n = NIL_P(n_value) ? 1 : NUM2INT(n_value);
```
byroot, line 299: So if I read the documentation correctly, this define the interval of SQLite VM instructions after which the callback is called. 

I'm not sure how many instructions are executed for each query, but I suspect even the most basic query at least invoke 1 instruction per row it scans, as such `1` as a default seem extremely aggressive.

byroot (commented):
So, I'm no owner of this gem, and my interest for sqlite3 is rather limited, but if the goal is to implement a statement timeout, I'm not offering a Proc interface is the best.

To call back into Ruby you need to hold the GVL. Right now it's not a problem because the gem calls `sqlite3_step` with the GVL held. However there is demand for releasing the GVL when executing a query. I don't know if it will happen, but such API would render such change impossible.

If the main envisioned usage is a statement timeout, then it's probably better to implement it all in C so the overhead is much smaller and that it doesn't need to hold the GVL.

ext/sqlite3/database.c
fractaledmind: I had it set to 1000 originally, but then I thought maybe I should default to fine-grained progress reports. But I'm happy to change it back to 1000. The [Datasette](https://datasette.io/) project defaults to 1000 for its statement timeout implementation: https://github.com/simonw/datasette/blob/00e59ec461dc0150772b999c7cc15fcb9b507d58/datasette/utils/__init__.py#L182

fractaledmind:
> So, I'm no owner of this gem, and my interest for sqlite3 is rather limited, but if the goal is to implement a statement timeout, I'm not offering a Proc interface is the best.
> 
> To call back into Ruby you need to hold the GVL. Right now it's not a problem because the gem calls `sqlite3_step` with the GVL held. However there is demand for releasing the GVL when executing a query. I don't know if it will happen, but such API would render such change impossible.
> 
> If the main envisioned usage is a statement timeout, then it's probably better to implement it all in C so the overhead is much smaller and that it doesn't need to hold the GVL.

My thought was to provide layers of exposure. The `progress_handler` is a feature of SQLite that application or library devs might use in any number of ways. The way that _I_ want to use it, in the context of Rails applications, is to add a statement timeout. But I don't want to _only_ provide a public interface for a statement timeout and keep the lower-level SQLite feature hidden. It is a public SQLite interface, so I think it should be a public sqlite3-ruby interface. 

However, when we turn to the PR for the higher-level `statement_timeout` feature, I would love to implement that in C and not in Ruby. My draft PR implements it in Ruby because that was the implementation I could build in an hour at this stage in my journey to learning how to write C extension code that glues SQLite and Ruby together. But I'm down to challenge myself and see if I can't implement the `statement_timeout` feature in pure C for that PR. I do not believe that the implementation details of the `statement_timeout` feature should impact this PR, however, as I believe this low-level, public SQLite interface should be exposed to Ruby developers.

fractaledmind:
Another great use-case for a raw `progress_handler` is allowing apps/projects using a Fiber Scheduler to switch fibers during query execution

fractaledmind:
@tenderlove @flavorjones: I believe this is ready to go.

fractaledmind:
Just realized that in addition to switching fibers, this also unlocks the ability for apps to switch threads while queries are executing 👀

fractaledmind:
@flavorjones @tenderlove: Merged changes from `main`, resolved a couple Standard issues, and added a test demonstrating that the `progress_handler` can be used to improve concurrency. I'm still reading and thinking and experimenting on how best to take advantage of this in various contexts (primarily Rails), but I think that this could be a major feature which at least partially addresses issues like #287. 

Also, just to add some additional info, Simon Willison of Datasette did some research on approximately how long it takes for SQLite to execute 1,000 VM steps (https://github.com/simonw/datasette/issues/1679#issuecomment-1074454687) and the approximate answer is 0.1ms.

fractaledmind:
This is going to be a really nice feature when we merge it. I've been experimenting with using this to support better CPU saturation in multi-threaded environments and the results are quite nice.

noteflakes:
@fractaledmind you might want to look at how this is done in [Extralite](https://github.com/digital-fabric/extralite/pull/68). There are some nuances that are important to keep in mind. The basic problem is that the SQLite VM instruction counter is reset for each query, so depending on the `N` parameter passed to [sqlite3_progress_handler](https://sqlite.org/c3ref/progress_handler.html) you'll might not get any calls to your progress handler:

```c
// register progress handler and invoke it every 10 VM instructions
sqlite3_progress_handler(db, 10, my_progress_handler, NULL);
// running the following query repeatedly will never invoke the progress handler
while (true) sqlite3_exec(db, "select 1", NULL, NULL, NULL);
```

On the other hand, if you set a `N` number too low you risk seriously hurting the performance of your queries, since your progress handler is called too often!

So what I did was to set the `N` parameter (which I call `tick`) passed to `sqlite3_progress_handler` to a low value, and then increment a counter in the C-based progress handler. When the counter reaches a certain threshold (which I call `period`), I invoke the Ruby progress handler:

```c
// from the Extralite code:
int Database_progress_handler(void *ptr) {
  Database_t *db = (Database_t *)ptr;
  db->progress_handler.tick_count += db->progress_handler.tick;
  if (db->progress_handler.tick_count < db->progress_handler.period)
    goto done;

  db->progress_handler.tick_count -= db->progress_handler.period;
  rb_funcall(db->progress_handler.proc, ID_call, 0);
done:
  return 0;
}
```

Also, note that the progress handler above always returns 0, which tells SQLite that the query can continue. Interrupting a query from the progress handler can be done by raising an exception, or by calling `Database#interrupt`, which also causes an exception to be raised. I believe this design fits in better with the Ruby exception handling system.

PR #459, opened 2024-01-08
Synchronize threads to remove "sleep 1".

masamitsu-murase:
I revived `test_busy_handler_outwait` because I think that this test covers a good use case regarding `busy_handler`.

Ruby 1.9 and later does not allow a thread to unlock a `Mutex` locked by another thread, so I used `Queue` to synchronize the main thread and another thread.
I also refined other test methods to remove `sleep 1`, which was used to synchronize threads.

I hope this pull request improves this project, but please feel free to ignore it if it is not so useful.

Regards,
Murase

### DIFF
--- a/test/test_integration_pending.rb
+++ b/test/test_integration_pending.rb
@@ -4,6 +4,39 @@ require 'thread'
 require 'benchmark'
 
 class TC_Integration_Pending < SQLite3::TestCase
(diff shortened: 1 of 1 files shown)


flavorjones:
MacOS failures are unrelated, merging. @masamitsu-murase Thank you!

Issue #410, opened 2023-09-27
Implement the `sqlite3_stmt_status` interface?


fractaledmind:
For developers looking to investigate whether or not their database schema is making an appropriate use of indexes, they need information about the actual usage of the database and its queries. SQLite does provide some historical information about "the performance characteristics of the prepared statements" via the `sqlite3_stmt_status` interface:

> Each prepared statement maintains various [SQLITE_STMTSTATUS counters](https://www.sqlite.org/c3ref/c_stmtstatus_counter.html) that measure the number of times it has performed specific operations. These counters can be used to monitor the performance characteristics of the prepared statements. For example, if the number of table steps greatly exceeds the number of table searches or result rows, that would tend to indicate that the prepared statement is using a full table scan rather than an index.
>
> This interface is used to retrieve and reset counter values from a [prepared statement](https://www.sqlite.org/c3ref/stmt.html). The first argument is the prepared statement object to be interrogated. The second argument is an integer code for a specific [SQLITE_STMTSTATUS counter](https://www.sqlite.org/c3ref/c_stmtstatus_counter.html) to be interrogated. The current value of the requested counter is returned. If the resetFlg is true, then the counter is reset to zero after this interface call returns.

from: https://www.sqlite.org/c3ref/stmt_status.html

A quick code search shows that this C function is not present in this code base, meaning that there is no way for Ruby developers to get at this information. This is valuable information. Could we implement a bridge to provide this information to the Ruby interface?

flavorjones:
@fractaledmind Thanks for the feature suggestion.

It would be really helpful if you would spend sometime designing the Ruby API that you wish existed, and would write (failing) tests that demonstrate the Ruby API you want. Demonstrating how this feature would be used in code is very valuable and would save me hours of time researching and designing an API myself. (For context, my OSS backlog feels a bit overwhelming to me at the moment.)

Once an API has been fleshed out, and I understand how you expect it to work, I may be able to quickly draft a C implementation for you to kick the tires on.

fractaledmind:
Absolutely. And I opened the issue just to put a flag to help me remember this thought. I will take full responsibility for exploring a Ruby interface and provide failing tests to back this interface. 

Your OSS work is valuable, but I never want you to feel obligated to anything from me. Let me finish this message by saying explicitly, I am very grateful for your presence in the Ruby ecosystem and I have found you to be a truly top-tier OSS maintainer and contributor. Thank you for everything you do

[...]

PR #461, opened 2024-01-08
Implement sqlite3_stmt_status interface

fractaledmind:
Resolves https://github.com/sparklemotion/sqlite3-ruby/issues/410

This PR adds 9 instance methods to the `Statement` class:

* `fullscan_steps`
* `sorts`
* `autoindexes`
* `vm_steps`
* `reprepares`
* `runs`
* `filter_misses`
* `filter_hits`
* `memused`

Each method corresponds to one of the [counters](https://www.sqlite.org/c3ref/c_stmtstatus_counter.html) available to the [`stmt_status`](https://www.sqlite.org/c3ref/stmt_status.html) interface. 

I wasn't sure how best to name the methods. There were two paths I considered: either mapping from the counter constant names or trying to find the clearest, most semantic name for the value. I decided to go with path one as this should make searching for methods from SQLite reference easier. We could add method aliases if we want to add more semantic method names.

### DIFF
--- a/ext/sqlite3/statement.c
+++ b/ext/sqlite3/statement.c
@@ -415,7 +415,8 @@ bind_parameter_count(VALUE self)
  *
  * Return the number of times that SQLite has stepped forward in a table as part of a full table scan
  */
-static VALUE fullscan_steps(VALUE self)
+static VALUE
+fullscan_steps(VALUE self)
 {
     sqlite3StmtRubyPtr ctx;
     TypedData_Get_Struct(self, sqlite3StmtRuby, &statement_type, ctx);
@@ -428,7 +429,8 @@ static VALUE fullscan_steps(VALUE self)
  *
  * Return the number of sort operations that have occurred
  */
-static VALUE sorts(VALUE self)
+static VALUE
+sorts(VALUE self)
 {
     sqlite3StmtRubyPtr ctx;
     TypedData_Get_Struct(self, sqlite3StmtRuby, &statement_type, ctx);
@@ -441,7 +443,8 @@ static VALUE sorts(VALUE self)
  *
  * Return the number of rows inserted into transient indices that were created automatically in order to help joins run faster
  */
-static VALUE autoindexes(VALUE self)
+static VALUE
+autoindexes(VALUE self)
 {
     sqlite3StmtRubyPtr ctx;
     TypedData_Get_Struct(self, sqlite3StmtRuby, &statement_type, ctx);
@@ -454,7 +457,8 @@ static VALUE autoindexes(VALUE self)
  *
  * Return the number of virtual machine operations executed by the prepared statement
  */
-static VALUE vm_steps(VALUE self)
+static VALUE
+vm_steps(VALUE self)
 {
     sqlite3StmtRubyPtr ctx;
     TypedData_Get_Struct(self, sqlite3StmtRuby, &statement_type, ctx);
@@ -463,11 +467,13 @@ static VALUE vm_steps(VALUE self)
     return INT2NUM(sqlite3_stmt_status(ctx->st, SQLITE_STMTSTATUS_VM_STEP, 0));
 }
 
+#ifdef SQLITE_STMTSTATUS_REPREPARE
 /* call-seq: stmt.reprepares
  *
  * Return the number of times that the prepare statement has been automatically regenerated due to schema changes or changes to bound parameters that might affect the query plan.
  */
-static VALUE reprepares(VALUE self)
+static VALUE
+reprepares(VALUE self)
 {
     sqlite3StmtRubyPtr ctx;
     TypedData_Get_Struct(self, sqlite3StmtRuby, &statement_type, ctx);
@@ -475,12 +481,15 @@ static VALUE reprepares(VALUE self)
 
     return INT2NUM(sqlite3_stmt_status(ctx->st, SQLITE_STMTSTATUS_REPREPARE, 0));
 }
+#endif
 
+#ifdef SQLITE_STMTSTATUS_RUN
 /* call-seq: stmt.runs
  *
  * Return the number of times that the prepared statement has been run
  */
-static VALUE runs(VALUE self)
+static VALUE
+runs(VALUE self)
 {
     sqlite3StmtRubyPtr ctx;
     TypedData_Get_Struct(self, sqlite3StmtRuby, &statement_type, ctx);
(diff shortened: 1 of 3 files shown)


fractaledmind:
It looks like `system` installations don't have every statement counter constant. Some guidance on how best to account for undefined constants would be appreciated. Should I simply check and not define the method? Should I define the method, but have the body raise something like a `NoMethodError`? Should I do something else entirely?

fractaledmind:
It seems `ubuntu-latest, 3.3, system` runs SQLite v3.37.2. Bloom filters and thus the `SQLITE_STMTSTATUS_FILTER_HIT` and `SQLITE_STMTSTATUS_FILTER_MISS` counters were added in v3.38.0.

`sqlcipher` runs SQLite v3.15.2. That version is missing `SQLITE_STMTSTATUS_REPREPARE`, `SQLITE_STMTSTATUS_RUN`, `SQLITE_STMTSTATUS_FILTER_MISS`, `SQLITE_STMTSTATUS_FILTER_HIT`, and `SQLITE_STMTSTATUS_MEMUSED`. I have made each of those corresponding functions optionally defined.

fractaledmind:
I don't know what it means when GitHub Actions don't have any output, they just show a box that says "This job failed", but I know that there were a number of actions running concurrently last night from other PRs, so maybe it timed out.

@flavorjones @tenderlove: If you re-run CI, I expect everything to pass

flavorjones:
Hey! Thanks for opening this up. The code as written seems great.

I have a question about the API design, though: as sqlite continues to add new stats, we'll be adding new methods each time, which feels strange to me. Some existing APIs that have a similar purpose provide a single method that returns a hash of statistics, for example [`GC.stat`](https://docs.ruby-lang.org/en/master/GC.html#method-c-stat).

Looking at the implementation of `sqlite3_stmt_status` we can see that all the stats except `MEMUSED` only require some pointer dereferences, and so are going to be fast and non-blocking; and `MEMUSED` acquires a mutex on the database:

```c
int sqlite3_stmt_status(sqlite3_stmt *pStmt, int op, int resetFlag){
  Vdbe *pVdbe = (Vdbe*)pStmt;
  u32 v;
  // ... snip ... 
  if( op==SQLITE_STMTSTATUS_MEMUSED ){
    sqlite3 *db = pVdbe->db;
    sqlite3_mutex_enter(db->mutex);
    v = 0;
    db->pnBytesFreed = (int*)&v;
    assert( db->lookaside.pEnd==db->lookaside.pTrueEnd );
    db->lookaside.pEnd = db->lookaside.pStart;
    sqlite3VdbeDelete(pVdbe);
    db->pnBytesFreed = 0;
    db->lookaside.pEnd = db->lookaside.pTrueEnd;
    sqlite3_mutex_leave(db->mutex);
  }else{
    v = pVdbe->aCounter[op];
    if( resetFlag ) pVdbe->aCounter[op] = 0;
  }
  return (int)v;
}
```

What do you think of making this two methods:

- `Statement#stat` that returns a hash containing all the supported statistics except `MEMUSED`
- `Statement#memused` that returns an int

?

fractaledmind:
@flavorjones: Done in https://github.com/sparklemotion/sqlite3-ruby/pull/461/commits/b9bbeaa7f1e2df83bf0694d096e0ae41a7262761. 

I implemented it like `GC.stat`, so you can pass no argument and just get a hash of stats back, or you can pass a hash, which will have the stats added to it, or you can pass a symbol for a particular stat and only that stat will be computed and returned to you.

In order to handle the 0 or 1 arguments, I followed the pattern from `db_filename`; that is, I defined a C method called `stmt_stat` which requires 1 argument that can be `nil`, `Hash`, or `Symbol`. I then define a `Statement#stat` method in Ruby which use a default argument to allow no argument to be passed.

For tests, I test that `stmt.stat` returns a Hash and then I just adapted the original tests to test that `stmt.stat(key)` returns the correct calculation. To handle some stats not being present in some builds, each stat test calls `stmt.stat` to get the hash of stats to see if the key is present.

I also added documentation for the `Statement#stat` method defined in Ruby. 

I believe this covers everything, but let me know if there is something you still want to evolve.

ext/sqlite3/statement.c
```diff
@@ -447,9 +587,10 @@ init_sqlite3_statement(void)
     rb_define_method(cSqlite3Statement, "column_name", column_name, 1);
     rb_define_method(cSqlite3Statement, "column_decltype", column_decltype, 1);
     rb_define_method(cSqlite3Statement, "bind_parameter_count", bind_parameter_count, 0);
-    rb_define_private_method(cSqlite3Statement, "prepare", prepare, 2);
-
-#ifdef HAVE_SQLITE3_COLUMN_DATABASE_NAME
-    rb_define_method(cSqlite3Statement, "database_name", database_name, 1);
+    rb_define_method(cSqlite3Statement, "stmt_stat", stmt_stat, 1);
```
tenderlove, line 590: I think we should probably make this private, especially if we have the Ruby level `stat` method.

lib/sqlite3/statement.rb
```diff
@@ -145,6 +145,29 @@ def must_be_open! # :nodoc:
...
+    #   - +reprepares+: the number of times that the prepare statement has been
+    # automatically regenerated due to schema changes or changes to bound
+    # parameters that might affect the query plan
+    #   - +runs+: the number of times that the prepared statement has been run
+    #   - +filter_misses+: the number of times that the Bloom filter returned
+    # a find, and thus the join step had to be processed as normal
+    #   - +filter_hits+: the number of times that a join step was bypassed
+    # because a Bloom filter returned not-found
+    def stat hash_or_key = nil
+      stmt_stat hash_or_key
```
tenderlove, line 168: Just a suggestion: if we implement 2 private methods in C then maybe we could pull branching logic in to Ruby.

For example:

```ruby
def stat key = nil
  if key
    stat_for(key)
  else
    stats_as_hash
  end
end
```

Where `stats_as_hash` and `stat_for_key` are private methods implemented in C.  WDYT?

tenderlove (changes requested):
I think we should make the C function `stmt_stat` private. I left one suggestion, but it's not required.

lib/sqlite3/statement.rb
fractaledmind: I like that. I started off by just trying to understand how the `GC.stat` method was implemented and oriented myself around that. However, once I realized I needed an optional argument and couldn't figure out how to do that in C and so make the Ruby method, I agree that simplifying the C code to just have two separate methods makes better sense. Will do that now

ext/sqlite3/statement.c
fractaledmind: Done in 089d00e

lib/sqlite3/statement.rb
fractaledmind: Done in 089d00e

tenderlove (approved):

PR #465, opened 2024-01-08
style: introduce rubocop, make some sensible changes

flavorjones:
- introduce rubocop
  - and add all violations to a TODO list so we can clean them up at our own pace
- cleanup whitespace violations
- cleanup inclusive language violations
  - add `SQLite3::Constants::Open::SUPER_JOURNAL` which is an alias for the deprecated `MASTER_JOURNAL`

Note that I've chosen to use the `Shopify/ruby-style-guide` rubocop config, because I like it and it's what I'm already using for Nokogiri. I don't feel strongly about it.


flavorjones:
Chatted with @tenderlove real quick and he convinced me to try standard instead, so expect this to get updated.

PR #467, opened 2024-01-09
introduce standard formatting and linting

flavorjones:
- introduce standard
- autocorrect safe and unsafe corrections, and apply a few manual fixes
- add `SQLite3::Constants::Open::SUPER_JOURNAL` which is an alias for the deprecated `MASTER_JOURNAL`
- default rake task: add `rubocop` and drop `clobber`
- fix a couple of tests


tenderlove (approved):
This looks great! Thank you for doing this.

PR #468, opened 2024-01-09
Use a factory method to eliminate "results as hash" check

tenderlove:
"Results as hash" should be set on the database, so we should only need to check it once when we first start the query.  This patch pulls result set construction in to a factory method on the database object.  Doing this eliminates a "translate to hash" check that is performed on every row fetch.


fractaledmind:
Love seeing these small details improvements. I'm learning a fair bit just from reading thru the recent PRs. Looking forward to v2.

PR #473, opened 2024-01-10
Remove the ResultSet object

tenderlove:
This commit removes the ResultSet object but also makes PreparedStatement objects quack the same as ResultSet.


tenderlove:
I'm not sure about this PR.  It seems like we shouldn't need the `ResultSet` object, but this patch impacts the behavior of `prepare` when someone passes `:results_as_hash => true` to the database.  This PR breaks ActiveRecord, so I'm going to port the code ActiveRecord is calling to a SQLite3 test so we can prevent regressions.

FWIW, Rails is [defaulting to `:results_as_hash => true`](https://github.com/rails/rails/blob/c66b655bf8948622722c24e6455b9b41febbe3a6/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L127), and I don't think that's a good idea.  I will also try to change Rails to not do that (if possible).

tenderlove:
I'm going to close this. We can revisit it in the future, but I don't think it's necessary for 2.0.0.

flavorjones (approved):
I really like this change! This is a nicer API, though I think we will have to copyedit some of the docstrings that are still referencing `ResultSet`.

Did you consider having `#each` et al implicitly call `#execute` if no results are loaded? That would be a further simplification which would allow something like

```ruby
db = SQLite3::Database.new(path)
db.prepare("select * from foo").each_row do |row|
  # ...
end
```

but I guess it's not that hard to jam an "execute" in that chain, either.

I searched github for instances of `SQLite3::Resultset.new` and saw that litedb breaks with this change for uninteresting reasons (it reimplements `#prepare`), but otherwise I think this is a safe change to make? We can open a PR with litestack.

But as an aside, we should probably add an integration test for litestack to the CI suite. Unfortunately it looks broken right now.

flavorjones:
I wrote my comment before I read yours. I'm fine with closing it.

tenderlove:
> But as an aside, we should probably add an integration test for litestack to the CI suite. Unfortunately it looks broken right now.

💯

PR #475, opened 2024-01-11
Implement GVL-releasing, backing-off busy_timeout

fractaledmind:
After getting a tad more experienced with the C portion of this codebase, I decided to try and see if I could implement the logic of SQLite's `busy_timeout` (with its backoff logic) in the extension such that we could release the GVL while sleeping. You can find SQLite's implementation of `sqlite3_busy_timeout` [here](https://github.com/sqlite/sqlite/blob/e5b2132df69e43ebe51df3a400dc2aba24f5fd29/src/main.c#L1835) and the default callback functionality [here](https://github.com/sqlite/sqlite/blob/e5b2132df69e43ebe51df3a400dc2aba24f5fd29/src/main.c#L1713). 

I think this is a good start, but I am stumped because when I try to compile, I get this error:
```
error: implicit declaration of function 'rb_thread_call_without_gvl' is invalid in C99
```

I'm not certain why this method isn't available, but how can I sleep and release the GVL here in C-land? Any help on this would be greatly appreciated. (note: I see from CI that this method is only missing in `darwin` environments. Interesting)

cc @suwyn who originally challenged me to do this in C-land.

### DIFF
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -559,6 +559,38 @@ set_authorizer(VALUE self, VALUE authorizer)
     return self;
 }
 
+static int
+rb_sqlite3_default_busy_callback(void *context, int count)
+{
+    sqlite3RubyPtr ctx = (sqlite3RubyPtr)context;
+    int timeout = ctx->busy_timeout;
+    int delay;
+    static const uint8_t delays[] =
+      { 1, 2, 5, 10, 15, 20, 25, 25,  25,  50,  50, 100 };
+    static const uint8_t totals[] =
+      { 0, 1, 3,  8, 18, 33, 53, 78, 103, 128, 178, 228 };
+    static const int n_delay = 12;
+    int prior;
+
+    assert( count>=0 );
+    if( count < n_delay ){
+      delay = delays[count];
+      prior = totals[count];
+    }else{
+      delay = delays[n_delay-1];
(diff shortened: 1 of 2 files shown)


fractaledmind:
@byroot: I know that at this point you are spending a meaningful portion of your working day (and maybe outside of work day) helping me learn C in the context of this library. I don't want to be a burden or take advantage of your generosity, so feel free to ignore this until you are totally free. However, if at some point you could help me understand how to release the GVL while sleeping, I would be greatly appreciative. 

Also, do you think we gain notable performance with such a C implementation of a timeout as opposed to the Ruby implementation recently merged?

byroot:
> when I try to compile, I get this error:

```
$ rg -F rb_thread_call_without_gvl ~/.rbenv/versions/3.3.0/include/ruby-3.3.0/ruby
/Users/byroot/.rbenv/versions/3.3.0/include/ruby-3.3.0/ruby/thread.h
89: *                 of a rb_thread_call_without_gvl()'s callback.
109: * Unlike rb_thread_call_without_gvl2()  this function  also reacts  to signals
130:void *rb_thread_call_without_gvl(void *(*func)(void *), void *data1,
```

`rb_thread_call_without_gvl` is defined in `ruby/thread.h` which isn't included by `ruby.h`, so you need `# include "ruby.h"`.

fractaledmind:
Understood. Will close for now then.

PR #480, opened 2024-01-23
Freeze results that come from the database

tenderlove:
Freeze strings and row records so that we can more easily share results among Ractors

Silly example, but here's an example of using database results plus Ractors to make a parallel producer-consumer pattern:

```ruby
require "sqlite3"
require "securerandom"

def fib n
  if n < 2
    n
  else
    fib(n-2) + fib(n-1)
  end
end

N = 10

# Make a Ractor safe queue that supports N workers
queue = Ractor.new do
  while (work = Ractor.receive)
    Ractor.yield work
  end
  # Shutdown the workers
  N.times { Ractor.yield nil }
end

# Spin up workers
workers = N.times.map do
  Ractor.new(queue) do |q|
    while (work = q.take)
      name, i = work
      p "Ractor: #{Ractor.current} #{name}: #{fib(i)}"
    end
  end
end

db = SQLite3::Database.new ":memory:"
db.execute 'CREATE TABLE "fibs" (name string, fib int)'
stmt = db.prepare "INSERT INTO fibs (name, fib) VALUES (?, ?)"
5000.times { stmt.execute SecureRandom.hex, rand(30) }
stmt.close

stmt = db.prepare 'SELECT name, fib FROM fibs'

# Send work from the database to the queue
stmt.each { |row| queue.send row }

stmt.close

# Tell the queue we've sent all the work
queue.send nil

# Wait for workers to finish
workers.map(&:take)
```

### DIFF
--- a/ext/sqlite3/statement.c
+++ b/ext/sqlite3/statement.c
@@ -162,6 +162,7 @@ step(VALUE self)
                         if (internal_encoding) {
                             val = rb_str_export_to_enc(val, internal_encoding);
                         }
+                        rb_obj_freeze(val);
                     }
                     break;
                     case SQLITE_BLOB: {
@@ -169,6 +170,7 @@ step(VALUE self)
                                   (const char *)sqlite3_column_blob(stmt, i),
                                   (long)sqlite3_column_bytes(stmt, i)
                               );
+                        rb_obj_freeze(val);
                     }
                     break;
                     case SQLITE_NULL:
@@ -191,6 +193,8 @@ step(VALUE self)
             ctx->done_p = 0;
             CHECK(sqlite3_db_handle(ctx->st), value);
     }
+
+    rb_obj_freeze(list);
 
     return list;
 }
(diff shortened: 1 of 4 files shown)


tenderlove:
This change breaks one test in Rails, so I changed Rails: https://github.com/rails/rails/pull/50859

ext/sqlite3/statement.c
```diff
@@ -162,13 +162,15 @@ step(VALUE self)
                         if (internal_encoding) {
                             val = rb_str_export_to_enc(val, internal_encoding);
                         }
+                        rb_obj_freeze(val);
```
flavorjones, line 165: Hmm, this seems to break Rails's schema cache.

```
bin/test test/cases/connection_adapters/schema_cache_test.rb -n/test_gzip_dumps_identical/
```

Without this line, the test passes. With it, the test fails (in isolation).

I'm looking into it.

flavorjones: Context on that test: https://github.com/rails/rails/pull/49166

flavorjones: See https://github.com/rails/rails/pull/51581 for the exciting conclusion.

PR #481, opened 2024-01-24
Ensure the `Database#encoding` method works when results are "as hash"

tenderlove:
This is a regression test for Rails. Rails instantiates SQLite3 with `results_as_hash` set to `true` by default. I broke the encoding method when that value is set to true, so this patch fixes it.


flavorjones:
@tenderlove thanks -- I should be able to get to setting up downstream integration testing with rails this week.

PR #486, opened 2024-01-24
Intern column names so we always get the same string

tenderlove:
When we get column names back from the database, it's very common to always return the same strings. This patch uses Ruby's "interned string" API so that we're always getting the same string objects back from the database.

Fixes: #155

### DIFF
--- a/ext/sqlite3/statement.c
+++ b/ext/sqlite3/statement.c
@@ -364,6 +364,22 @@ column_count(VALUE self)
     return INT2NUM(sqlite3_column_count(ctx->st));
 }
 
+#if HAVE_RB_ENC_INTERNED_STR_CSTR
+static VALUE
+interned_utf8_cstr(const char * str)
+{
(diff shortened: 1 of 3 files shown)


ext/sqlite3/statement.c
```diff
@@ -364,6 +364,22 @@ column_count(VALUE self)
...
+interned_utf8_cstr(const char * str)
+{
+    return rb_enc_interned_str_cstr(str, rb_utf8_encoding());
+}
+#else
+static VALUE
+interned_utf8_cstr(const char * str)
+{
+    VALUE rb_str = rb_utf8_str_new_cstr(str);
+    rb_obj_freeze(rb_str);
```
eregon, line 378: Minor: this call is unnecessary, because `String#-@` always returns a frozen String.

flavorjones: Thanks, fixed.

PR #497, opened 2024-02-01
Add threadsafety documentation

### LINKED ISSUE #391: Gem is not threadsafe
After spending quite some time chasing an obscure bug in one of my employer's applications, I thought I'd share my experiences here so that other people might benefit from it.

Our code contained a singleton that used a prepared statement to query a SQLite DB in read-only mode. This led to strange problems like empty result sets for queries that were guaranteed to return at least one row.

For us, these issues only manifested when running the app in production with Puma (a multi-threaded web server).

Here is a simple reproducer:

```ruby
#!/usr/bin/env ruby

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'sqlite3', '~> 1.6.3'
end

# Ensure that we have a thread-safe SQLite library for this test
raise 'SQLite3 is not threadsafe' unless SQLite3.threadsafe?

db = SQLite3::Database.new ':memory:'
statement = db.prepare('SELECT :foo')

threads = (0...5).map do |i|
  Thread.new do
    pp "Thread #{i}: starting"

    1000.times do
      pp "Thread #{i}: lookup"

      result = statement.execute!(foo: 1)

      "Thread #{i}: error" if result.nil? || result.empty?
    end

    pp "Thread #{i}: done"
  end
end

threads.each(&:join)
```

This will eventually result in an exception like this:

```
#<Thread:0x00007ff4471c2c10 ./bug.rb:17 run> terminated with exception (report_on_exception is true):
/home/mz/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/sqlite3-1.6.3-x86_64-linux/lib/sqlite3/statement.rb:39:in `bind_param': bad parameter or other API misuse (SQLite3::MisuseException)
        from /home/mz/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/sqlite3-1.6.3-x86_64-linux/lib/sqlite3/statement.rb:39:in `block (2 levels) in bind_params'
        from /home/mz/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/sqlite3-1.6.3-x86_64-linux/lib/sqlite3/statement.rb:39:in `each'
        from /home/mz/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/sqlite3-1.6.3-x86_64-linux/lib/sqlite3/statement.rb:39:in `block in bind_params'
        from /home/mz/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/sqlite3-1.6.3-x86_64-linux/lib/sqlite3/statement.rb:37:in `each'
        from /home/mz/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/sqlite3-1.6.3-x86_64-linux/lib/sqlite3/statement.rb:37:in `bind_params'
        from /home/mz/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/sqlite3-1.6.3-x86_64-linux/lib/sqlite3/statement.rb:64:in `execute'
        from /home/mz/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/sqlite3-1.6.3-x86_64-linux/lib/sqlite3/statement.rb:88:in `execute!'
        from ./bug.rb:23:in `block (3 levels) in <main>'
        from ./bug.rb:20:in `times'
        from ./bug.rb:20:in `block (2 levels) in <main>'
```

In hindsight, this behaviour is not surprising me at all.

My concern is that it is very easy to introduce this kind of concurrency bug, even if the application code itself does not do any explicit threading.

Judging by the discussions in some other issues (e.g. #217, #287), I think that the README should contain a prominent warning about using this gem in multi-threaded code; even if `libsqlite3.so` itself has been [compiled with `-DSQLITE_THREADSAFE=2`](https://www.sqlite.org/threadsafe.html).

tenderlove:
Fixes #391

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,78 @@ end
 # => ["Jane", "me@janedoe.com", "A", "http://blog.janedoe.com"]
 ```
 
+## Thread Safety
+
+When `SQLite3.threadsafe?` returns `true`, then SQLite3 has been compiled to
+support running in a multithreaded environment.  However, this doesn't mean
+that all classes in the SQLite3 gem can be considered "thread safe".
+
+When `SQLite3.threadsafe?` returns `true`, it is safe to share only
+`SQLite3::Database` instances among threads without providing your own locking
+mechanism.  For example, the following code is fine because only the database
+instance is shared among threads:
+
+```ruby
+require 'sqlite3'
+
+db = SQLite3::Database.new ":memory:"
+
+latch = Queue.new
+
+ts = 10.times.map {
+  Thread.new {
+    latch.pop
+    db.execute "SELECT '#{Thread.current.inspect}'"
+  }
+}
+10.times { latch << nil }
+
+p ts.map(&:value)
+```
+
+Other instances can be shared among threads, but they require that you provide
+your own locking for thread safety.  For example, `SQLite3::Statement` objects
+(prepared statements) are mutable, so applications must take care to add
+appropriate locks to avoid data race conditions when sharing these objects
+among threads.
+
+Lets rewrite the above example but use a prepared statement and safely share
+the prepared statement among threads:
+
+```ruby
+db = SQLite3::Database.new ":memory:"
+
+# Prepare a statement
+stmt = db.prepare "SELECT :inspect"
+stmt_lock = Mutex.new
+
+latch = Queue.new
+
+ts = 10.times.map {
+  Thread.new {
+    latch.pop
+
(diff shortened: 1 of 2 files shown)


flavorjones (approved):

ioquatix:
Nice documentation!

Issue #489, opened 2024-01-26
occasional ruby_memcheck CI failures


flavorjones:
I've seen ~4 of these in the past week while we've been running a lot of CI jobs

```
77,768 (80 direct, 77,688 indirect) bytes in 1 blocks are definitely lost in loss record 162 of 162
  malloc (at /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
 *sqlite3MemMalloc (at /home/runner/work/sqlite3-ruby/sqlite3-ruby/lib/sqlite3/sqlite3_native.so)
 *sqlite3Malloc (at /home/runner/work/sqlite3-ruby/sqlite3-ruby/lib/sqlite3/sqlite3_native.so)
 *sqlite3BtreeOpen (at /home/runner/work/sqlite3-ruby/sqlite3-ruby/lib/sqlite3/sqlite3_native.so)
 *openDatabase (at /home/runner/work/sqlite3-ruby/sqlite3-ruby/lib/sqlite3/sqlite3_native.so)
 *rb_sqlite3_open_v2 (database.c:92)
  vm_call_cfunc_with_frame_ (vm_insnhelper.c:3490)
  vm_call_cfunc_with_frame (vm_insnhelper.c:3518)
  vm_sendish (vm_insnhelper.c:5581)
```

or 

```
77,272 (40 direct, 77,232 indirect) bytes in 1 blocks are definitely lost in loss record 163 of 163
  malloc (at /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
 *sqlite3MemMalloc (at /home/runner/work/sqlite3-ruby/sqlite3-ruby/lib/sqlite3/sqlite3_native.so)
 *sqlite3Malloc (at /home/runner/work/sqlite3-ruby/sqlite3-ruby/lib/sqlite3/sqlite3_native.so)
 *sqlite3HashInsert (at /home/runner/work/sqlite3-ruby/sqlite3-ruby/lib/sqlite3/sqlite3_native.so)
 *sqlite3FindFunction (at /home/runner/work/sqlite3-ruby/sqlite3-ruby/lib/sqlite3/sqlite3_native.so)
 *sqlite3CreateFunc (at /home/runner/work/sqlite3-ruby/sqlite3-ruby/lib/sqlite3/sqlite3_native.so)
[...]

PR #500, opened 2024-02-02
fix: expanded sql leak

flavorjones:
cc @tenderlove 

This actually was caught by the test suite in https://github.com/sparklemotion/sqlite3-ruby/actions/runs/7745352446/job/21121203412#step:6:82 but I missed it because we've been having some flaky failures of the valgrind/memcheck suite (see https://github.com/sparklemotion/sqlite3-ruby/issues/489).


tenderlove:
Oh wow, thank you. I didn't realize you had to free the string, but I guess it makes sense. 🤦🏻‍♀️

tenderlove (approved):

PR #508, opened 2024-02-28
Database#transaction returns block result

### LINKED ISSUE #113: Why does Database#transaction always return true?
Database#transaction always return true.

This doesn't strike me as very helpful behaviour - it doesn't allow you to simply pull the result of the last SQL statement executed in the transaction as the result of the transaction

I have suggest the following modification (which I have submitted in a pull request)
Database#transaction should return
- the return value of the block (if not nil)
- otherwise true if transaction committed

I will submit a patch against this issue :-)

alexcwatt:
Closes #113.


flavorjones (approved):

flavorjones:
Awesome! Thank you so much for contributing this improvement.

PR #509, opened 2024-03-01
Remove FunctionProxy#count

alexcwatt:
This code depends on a SQLite API which is deprecated, `sqlite3_aggregate_count` ([docs](https://www.sqlite.org/c3ref/aggregate_count.html)). In addition, the present implementation is broken (`@driver` is always nil). Related: #164.

## Note to reviewers

[This check](https://github.com/sparklemotion/sqlite3-ruby/actions/runs/8106559146/job/22156728792?pr=509) failed and appears unrelated to this code change.

### DIFF
--- a/lib/sqlite3/database.rb
+++ b/lib/sqlite3/database.rb
@@ -725,13 +725,6 @@ module SQLite3
         @driver.result_error(@func, error.to_s, -1)
       end
 
-      # (Only available to aggregate functions.) Returns the number of rows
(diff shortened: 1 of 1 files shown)


flavorjones:
@alexcwatt Thanks! I'll look at the CI failure.

flavorjones:
CI was a flake. This looks good to me, after looking at the history it seems like this should have been removed in aaa74ebd back in (gulp) 2010.

Issue #243, opened 2018-12-30
execute_batch returns the result of the last statement


donaldong:
Currently `execute_batch` always return `nil`.
1) Would it be better to return the result for the last statement instead?
2) Would it make sense to accept a list of statements as well? For example:
```
db.execute_batch [
  "PRAGMA empty_result_callbacks=1",
  "CREATE TABLE t1(a,b)",
  "CREATE INDEX t1i3 ON t1(a,b)",
  "INSERT INTO t1 VALUES(1, 2)",
  "SELECT * FROM t1",
]
```
or simply `db.execute [...]`?

tenderlove:
Yes, I think both of those things would be fine.

PR #512, opened 2024-03-02
execute_batch returns result of last statement

alexcwatt:
This is part of what's requested in https://github.com/sparklemotion/sqlite3-ruby/issues/243. (The other part is to take an array of statements, but that is not implemented here.)


flavorjones:
Looks great, thank you!

PR #520, opened 2024-04-15
Add SQLite3.status to call sqlite3_status

wjlroe:
This function is useful for understanding the runtime performance of sqlite.

This is a very thin layer on top of the sqlite3 library's function, so it allows full usage of it from Ruby.

I've tested with `./bin/build-gems` and it built fine and the tests passed.


flavorjones:
@wjlroe Thank you for this contribution!

I've added a commit with some documentation for the method, and tweaked the test, but otherwise this seems great. Thank you!

flavorjones:
I've also updated `SQLite3.status` to make its second argument optional.

wjlroe:
@flavorjones Thanks for improving this! It looks like some of the tests are failing because it's returning 0 sometimes, which I was a bit unsure about when I wrote that test - that test might have to do something via sqlite before enquiring about memory to ensure it's non-zero - what do you think?

flavorjones:
@wjlroe Yep, I updated the tests to test for `>= 0` so they should be OK now!

PR #526, opened 2024-04-19
fix: Database#execute_batch properly handles non-ASCII strings

### LINKED ISSUE #524: SQLite3::Database#execute_batch fails with non-ASCII characters
Before v2.0.0, the following used to work:

```ruby
require 'sqlite3'

db = SQLite3::Database.new ':memory:'

db.execute_batch %{
  CREATE TABLE test(t TEXT);
  INSERT INTO test VALUES ('héhé');
}
```

But now fails with:

```
/usr/home/mathieu/.gem/ruby/3.2/gems/sqlite3-2.0.0/lib/sqlite3/statement.rb:31:in `encode': "\xC3" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
        from /usr/home/mathieu/.gem/ruby/3.2/gems/sqlite3-2.0.0/lib/sqlite3/statement.rb:31:in `initialize'
        from /usr/home/mathieu/.gem/ruby/3.2/gems/sqlite3-2.0.0/lib/sqlite3/database.rb:166:in `new'
        from /usr/home/mathieu/.gem/ruby/3.2/gems/sqlite3-2.0.0/lib/sqlite3/database.rb:166:in `prepare'
        from /usr/home/mathieu/.gem/ruby/3.2/gems/sqlite3-2.0.0/lib/sqlite3/database.rb:250:in `execute_batch'
        from ./test.rb:5:in `<main>'

```

flavorjones:
Fix a regression in v2.0.0 that caused `Database#execute_batch` to raise an encoding exception when passed some non-ASCII strings. As a result of this fix, `Database#prepare` now ensures the "remainder" string will always be encoded as UTF-8.

Closes #524

### DIFF
--- a/ext/sqlite3/statement.c
+++ b/ext/sqlite3/statement.c
@@ -72,7 +72,7 @@ prepare(VALUE self, VALUE db, VALUE sql)
     CHECK(db_ctx->db, status);
     timespecclear(&db_ctx->stmt_deadline);
 
-    return rb_str_new2(tail);
+    return rb_utf8_str_new_cstr(tail);
 }
 
 /* call-seq: stmt.close
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - Raise an exception if `Database#execute`, `#execute_batch`, or `#query` are passed multiple bind parameters that are not in an Array. In v2.0.0 these methods would silently swallow additional arguments. [#527] @flavorjones
+- Fixed a regression in v2.0.0 that caused `Database#execute_batch` to raise an encoding exception when passed some non-ascii strings. As a result of this fix, `Database#prepare` now ensures the "remainder" string will always be encoded as UTF-8. [#524] @flavorjones
 
 
 ## 2.0.0 / 2024-04-17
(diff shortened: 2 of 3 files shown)


ext/sqlite3/statement.c
```diff
@@ -72,7 +72,7 @@ prepare(VALUE self, VALUE db, VALUE sql)
     CHECK(db_ctx->db, status);
     timespecclear(&db_ctx->stmt_deadline);
 
-    return rb_str_new2(tail);
+    return rb_utf8_str_new_cstr(tail);
```
Math2, line 75: Would it be better to reuse the encoding of the original string?

flavorjones: Regardless of the original encoding, the query ends up encoded as UTF-8 in `Statement#initialize` and the remainder string being returned from `sqlite_prepare` is also UTF-8.

This call isn't re-encoding anything, it's just telling Ruby that the C string is in fact UTF-8 (the default is otherwise ASCII-8BIT which is what led to the problem you reported).

Math2: Ah, I see. Alright, thanks for fixing this!

PR #532, opened 2024-04-25
test: improve the "busy handler timeout" GVL test

flavorjones:
Following the pattern from #459 let's speed the GVL tests up


flavorjones:
All tests went green. Re-running just to check for flakes.

PR #536, opened 2024-05-23
dep: update vendored sqlite3 to 3.46.0

flavorjones:
https://sqlite.org/releaselog/3_46_0.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # sqlite3-ruby Changelog
 
+## next / unreleased
+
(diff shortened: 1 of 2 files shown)


flavorjones:
The failing downstream test is because activerecord's tests are overly-specific about the expected error messages. Addressed in https://github.com/rails/rails/pull/51891

flavorjones:
downstream test is green after the rails PR was merged: https://github.com/sparklemotion/sqlite3-ruby/actions/runs/9210959257/job/25341559773

PR #554, opened 2024-08-17
Provide the SQL statement context 

fractaledmind:
for errors that occur when preparing a SQL statement

@flavorjones @tenderlove @byroot 

This is a proof-of-concept PR to explore providing better context for exceptions raised. To begin, I focused on exceptions raised when attempting to prepare a statement. Using the gem over the years, there are other contexts that I found the error messages lacking, but I need to recall them and work on PoCs for each. But I wanted to open this to start the conversation around improving error messages with context.

### DIFF
--- a/ext/sqlite3/exception.c
+++ b/ext/sqlite3/exception.c
@@ -115,3 +115,116 @@ rb_sqlite3_raise_msg(sqlite3 *db, int status, const char *msg)
     rb_iv_set(exception, "@code", INT2FIX(status));
     rb_exc_raise(exception);
 }
+
+void
+rb_sqlite3_raise_with_sql(sqlite3 *db, int status, const char *sql)
+{
+    VALUE klass = Qnil;
+    VALUE error_message = Qnil;
+    const char* sqlite_error_msg = sqlite3_errmsg(db);
+    int error_offset = sqlite3_error_offset(db);
+
+    /* Consider only lower 8 bits, to work correctly when
+       extended result codes are enabled. */
+    switch (status & 0xff) {
+        case SQLITE_OK:
+            return;
+            break;
+        case SQLITE_ERROR:
+            klass = rb_path2class("SQLite3::SQLException");
+            break;
+        case SQLITE_INTERNAL:
+            klass = rb_path2class("SQLite3::InternalException");
+            break;
+        case SQLITE_PERM:
+            klass = rb_path2class("SQLite3::PermissionException");
+            break;
+        case SQLITE_ABORT:
+            klass = rb_path2class("SQLite3::AbortException");
+            break;
+        case SQLITE_BUSY:
+            klass = rb_path2class("SQLite3::BusyException");
+            break;
+        case SQLITE_LOCKED:
+            klass = rb_path2class("SQLite3::LockedException");
+            break;
+        case SQLITE_NOMEM:
+            klass = rb_path2class("SQLite3::MemoryException");
+            break;
+        case SQLITE_READONLY:
+            klass = rb_path2class("SQLite3::ReadOnlyException");
+            break;
+        case SQLITE_INTERRUPT:
+            klass = rb_path2class("SQLite3::InterruptException");
+            break;
+        case SQLITE_IOERR:
+            klass = rb_path2class("SQLite3::IOException");
+            break;
+        case SQLITE_CORRUPT:
+            klass = rb_path2class("SQLite3::CorruptException");
+            break;
+        case SQLITE_NOTFOUND:
+            klass = rb_path2class("SQLite3::NotFoundException");
+            break;
+        case SQLITE_FULL:
+            klass = rb_path2class("SQLite3::FullException");
+            break;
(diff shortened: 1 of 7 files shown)


ext/sqlite3/exception.c
```diff
@@ -115,3 +115,116 @@ rb_sqlite3_raise_msg(sqlite3 *db, int status, const char *msg)
...
+            klass = rb_path2class("SQLite3::NotADatabaseException");
+            break;
+        default:
+            klass = rb_path2class("SQLite3::Exception");
+    }
+
+    // Create a more detailed error message
+    if (error_offset >= 0 && sql) {
+        char *formatted_error = NULL;
+        asprintf(&formatted_error, "%s\n  %s\n  %*s^--- error here",
```
casperisfine, line 218: Would probably be better to use `rb_sprintf`, would also save you the `free`.

ext/sqlite3/exception.c
```diff
@@ -115,3 +115,116 @@ rb_sqlite3_raise_msg(sqlite3 *db, int status, const char *msg)
...
+        free(formatted_error);
+    } else {
+        error_message = rb_str_new2(sqlite_error_msg);
+    }
+
+    klass = rb_exc_new3(klass, error_message);
+    rb_iv_set(klass, "@code", INT2FIX(status));
+    rb_iv_set(klass, "@error_offset", INT2FIX(error_offset));
+    rb_exc_raise(klass);
+}
```
casperisfine, line 230:
```suggestion
}

```

test/test_integration.rb
```diff
@@ -102,9 +102,10 @@ def test_authorizer_silent
   end
 
   def test_prepare_invalid_syntax
-    assert_raise(SQLite3::SQLException) do
+    exception = assert_raise(SQLite3::SQLException) do
       @db.prepare "select from foo"
     end
+    assert_equal("near \"from\": syntax error\n  select from foo\n         ^--- error here", exception.message)
```
casperisfine, line 108: Heredocs are usually best for that sort of things.

```suggestion
    assert_equal(<<~MSG, exception.message)
      near "from": syntax error
        select from foo
                 ^--- error here
      MSG
```

ext/sqlite3/exception.c
```diff
@@ -115,3 +115,116 @@ rb_sqlite3_raise_msg(sqlite3 *db, int status, const char *msg)
...
+rb_sqlite3_raise_with_sql(sqlite3 *db, int status, const char *sql)
+{
+    VALUE klass = Qnil;
+    VALUE error_message = Qnil;
+    const char* sqlite_error_msg = sqlite3_errmsg(db);
+    int error_offset = sqlite3_error_offset(db);
+
+    /* Consider only lower 8 bits, to work correctly when
+       extended result codes are enabled. */
+    switch (status & 0xff) {
```
casperisfine, line 129: This should probably be exctraced in a function rather than copy pasted.

flavorjones: I strongly agree.

flavorjones (commented):
This seems like a great feature to add, if we can clean it up a little bit as suggested by @byroot.

Also please note there's a failing test where we run with an older version of sqlite that doesn't have `sqlite3_error_offset`. To deal with older versions you'll need to add a `have_func` in the extconf and then `#ifdef HAVE_SQLITE3_ERROR_OFFSET` in the code. Let me know if you want any help!

fractaledmind:
Thank you! I put this WIP PR out there after a quick "hacking session" while prepping talks. I'll get it cleaned up when I'm back from my grad conference adventures. 

But, if you have a clear vision and a hacking session to push this over the finish line, I have no sense of ownership over this PR. 

Either way, we can get this finished up in October at the latest.

ext/sqlite3/exception.c
```diff
@@ -115,3 +115,116 @@ rb_sqlite3_raise_msg(sqlite3 *db, int status, const char *msg)
...
+            break;
+        case SQLITE_FORMAT:
+            klass = rb_path2class("SQLite3::FormatException");
+            break;
+        case SQLITE_RANGE:
+            klass = rb_path2class("SQLite3::RangeException");
+            break;
+        case SQLITE_NOTADB:
+            klass = rb_path2class("SQLite3::NotADatabaseException");
+            break;
```
tenderlove, line 210: Would it be odd to expose `SQLITE_*` as private constants in Ruby, then call a Ruby method which raises the exception? Then we don't need to deal with the memory management below.

flavorjones:
I moved the sql error message handling into Ruby and made it backwards-compatible with older versions of sqlite. Would love a bit of review on this if you have a moment @fractaledmind @tenderlove @byroot.

byroot (approved):
Looks fine to me.

tenderlove (approved):
LGTM

fractaledmind:
Looks good to me too. The C code is indeed _much_ better, so thank you all for improving my PoC so considerably. I fixed the merge conflict in the CHANGELOG, so hopefully all the checks pass now.

PR #557, opened 2024-09-15
Use sqlite3_close_v2 to close databases.

flavorjones:
Close databases in a deferred manner if there are unclosed prepared statements. Previously closing a database while statements were open resulted in a `BusyException`.

See https://www.sqlite.org/c3ref/close.html for more context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # sqlite3-ruby Changelog
 
+## next / unreleased
+
+### Improved
+
+- Use `sqlite3_close_v2` to close databases in a deferred manner if there are unclosed prepared statements. Previously closing a database while statements were open resulted in a `BusyException`. See https://www.sqlite.org/c3ref/close.html for more context. @flavorjones
+
+
 ## 2.0.4 / 2024-08-13
 
 ### Dependencies
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -25,7 +25,7 @@ deallocate(void *ctx)
     sqlite3RubyPtr c = (sqlite3RubyPtr)ctx;
     sqlite3 *db     = c->db;
 
-    if (db) { sqlite3_close(db); }
+    if (db) { sqlite3_close_v2(db); }
     xfree(c);
 }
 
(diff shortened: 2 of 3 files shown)


tenderlove:
> Close databases in a deferred manner if there are unclosed prepared statements. Previously closing a database while statements were open resulted in a `BusyException`.

I know I'm late to the party (again, really sorry), but do we actually want to do this?  If someone closes a database while they have statements open, that is a bug, so I'd expect an exception.  We definitely need the "close without exception" behavior for fork safety, but could we just introduce a "force close" or something?  Maybe like `db.close(force: true)`

flavorjones:
@tenderlove Been thinking about this.

I think what you're saying is: maybe people will be silently leaking db connections because they're not being reminded to clean up their statements by a failing `Database#close` call?

I guess I get it, and I acknowledge it's a change in behavior from previous versions. But also how is this any different than anything else in Ruby? It was only a bug before because GC didn't always clean up databases and statements; now this is handled well.

Is there a specific use case you're worried about leaking in non-obvious ways?

tenderlove:
Just to close the loop on this (since we discussed it privately already):

> I think what you're saying is: maybe people will be silently leaking db connections because they're not being reminded to clean up their statements by a failing `Database#close` call?

I'm not worried that people will be silently leaking database connections. I'm worried that people will mistakenly leave statements in an "unfinished" state.  Previously, users would have to specifically close their statements before they could close the database, effectively forcing them to say "I'm done with this statement now" (even if the statement had more rows to read, etc) before being allowed to close the db.  After this patch, there's no feedback to users that they've **not** completed using a statement.

> I guess I get it, and I acknowledge it's a change in behavior from previous versions. But also how is this any different than anything else in Ruby?

Yep, totally agree with this. I think it's what we agreed upon privately, and I just want to make sure we document it publicly. 😄

> It was only a bug before because GC didn't always clean up databases and statements; now this is handled well.

I _was_ thinking we could do both: have good GC cleanup _and_ raise an exception if there are "dangling" statements.  I think the behavior we have in `main` now is good.  If _not_ raising an exception breaks someone's code, I believe we can put back the old behavior and have good GC handling. But lets not consider that until we find out it's actually a problem. I'm worried that I worried over nothing (if that makes sense).

PR #558, opened 2024-09-15
Open & writable database connections carried across fork() are automatically discarded in the child

flavorjones:
## Context

In August 2024, Andy Croll opened an issue[^issue] describing sqlite file corruption related to solid queue. After investigation, we were able to reproduce corruption under certain circumstances when forking a process with open sqlite databases.[^repro]

SQLite is known to not be fork-safe[^howto], so this was not entirely surprising though it was the first time your author had personally seen corruption in the wild. The corruption became much more likely after the sqlite3-ruby gem improved its memory management with respect to open statements[^gemleak] in v2.0.0.

Advice from upstream contributors[^advice] is, essentially: don't fork if you have open database connections. Or, if you have forked, don't call `sqlite3_close` on those connections and thereby leak some amount of memory in the child process. Neither of these options are ideal, see below.

## Decisions

1. Open writable database connections carried across a `fork()` will automatically be closed in the child process to mitigate the risk of corrupting the database file.
2. These connections will be incompletely closed ("discarded") which will result in a one-time memory leak in the child process.

First, the gem will register an "after fork" handler via `Process._fork` that will close any open writable database connections in the child process. This is a best-effort attempt to avoid corruption, but it is not guaranteed to prevent corruption in all cases. Any connections closed by this handler will also emit a warning to let users know what's happening.

Second, the sqlite3-ruby gem will store the ID of the process that opened each database connection. If, when a writable database is closed (either explicitly with `Database#close` or implicitly via GC or after-fork callback) the current process ID is different from the original process, then we "discard" the connection.

"Discard" here means:

- The `Database` object acts "closed", including returning `true` from `#closed?`.
- `sqlite3_close_v2` is not called on the object, because it is unsafe to do so per sqlite instructions[^howto]. As a result, some memory will be lost permanently (a one-time "memory leak").
- Open file descriptors associated with the database are closed.
- Any memory that can be freed safely is recovered.

Note that readonly databases are being treated as "fork safe" and are not affected by any of these changes.

## Consequences

The positive consequence is that we remove a potential cause of database corruption for applications that fork with active sqlite database connections.

The negative consequence is that, for each discarded connection, some memory will be permanently lost (leaked) in the child process.

### 1. Require applications to close database connections before forking.

This is the advice[^advice] given by the upstream maintainers of sqlite, and so was the first thing we tried to implement in Rails in [rails/rails#52931](https://github.com/rails/rails/pull/52931)[^before_fork]. That first simple implementation was not thread safe, however, and in order to make it thread-safe it would be necessary to pause all sqlite database activity, close the open connections, and then fork. At least one Rails core team member was not happy that this would interfere with database connections in the parent, and the complexity of a thread-safe solution seemed high, so this work was paused.

### 2. Memory arena

Sqlite offers a configuration option to specify custom memory functions for malloc et al. It seems possible that the sqlite3-ruby gem could implement a custom arena that would be used by sqlite so that in a new process, after forking, all the memory underlying the sqlite Ruby objects could be discarded in a single operation.

I think this approach is promising, but complex and risky. Sqlite is a complex library and uses shared memory in addition to the traditional heap. Would throwing away the heap memory (the arena) result in a segfault or other undefined behaviors or corruption? Determining the answer to that question feels expensive in and of itself, and any solution along these lines would not be supported by the sqlite authors. We can explore this space if the memory leak from discarded connections turns out to be a large source of pain.

## References

- [Database connections carried across fork() will not be fully closed by flavorjones · Pull Request #558 · sparklemotion/sqlite3-ruby](https://github.com/sparklemotion/sqlite3-ruby/pull/558)
- TODO rails pr implementing sqlite3adapter discard

## Footnotes

[^issue]: [SQLite queue database corruption · Issue #324 · rails/solid_queue](https://github.com/rails/solid_queue/issues/324)
[^repro]: [flavorjones/2024-09-13-sqlite-corruption: Temporary repo, reproduction of sqlite database corruption.](https://github.com/flavorjones/2024-09-13-sqlite-corruption)
[^howto]: [How To Corrupt An SQLite Database File: §2.6 Carrying an open database connection across a fork()](https://www.sqlite.org/howtocorrupt.html#_carrying_an_open_database_connection_across_a_fork_)
[^gemleak]: [Always call sqlite3_finalize in deallocate func by haileys · Pull Request #392 · sparklemotion/sqlite3-ruby](https://github.com/sparklemotion/sqlite3-ruby/pull/392)
[^advice]: [SQLite Forum: Correct way of carrying connections over forked processes](https://sqlite.org/forum/forumpost/1fa07728204567a0a136f442cb1c59e3117da96898b7fa3290b0063ae7f6f012)
[^before_fork]: [SQLite3Adapter: Ensure fork-safety by flavorjones · Pull Request #52931 · rails/rails](https://github.com/rails/rails/pull/52931#issuecomment-2351365601)
[^config]: [SQlite3 Configuration Options](https://www.sqlite.org/c3ref/c_config_covering_index_scan.html)


adr/2024-09-fork-safety.md
```diff
@@ -0,0 +1,57 @@
...
+The positive consequence is that frameworks like Rails can prevent database file corruption as a result of allowing forks by simply `discard`ing the inherited connections in the child process.
+
+The negative consequence is that, when choosing this option, some memory will be permanently lost (leaked) in the child process.
+
+
+## Alternatives considered.
+
+### 1. Require applications to close database connections before forking.
+
+This is the advice[^advice] given by the upstream maintainers of sqlite, and so was the first thing we tried to implement in Rails in [rails/rails#52931](https://github.com/rails/rails/pull/52931)[^before_fork]. That first simple implementation was not thread safe, however, and in order to make it thread-safe it would be necessary to pause all sqlite database activity, close the open connections, and then fork. At least one Rails core team member was not happy that this would interfere with database connections in the parent, and the complexity of a thread-safe solution seemed high, so this work was paused.
```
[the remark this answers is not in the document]
byroot: (sorry, just noticed now I queued the review comment instead of publishing it hours ago)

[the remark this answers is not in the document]
flavorjones: > have all the connection share a read-write lock

In what way would the connections share this? What operations would be behind the read lock?

I started implementing something like this -- a global shared/exclusive lock owned by the SQLite3Adapter class -- and then had `ConnectionPool#checkout` grab the shared lock. But there were still race conditions and closed connections being passed around somewhere. It didn't seem straightforward to me, but I'm not that familiar with connection management code or I may have misunderstood your suggestion.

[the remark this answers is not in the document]
flavorjones: If you want to take a look at what I started, it's here: https://github.com/flavorjones/rails/tree/flavorjones-sqlite3-fork-safety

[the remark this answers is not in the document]
byroot: > In what way would the connections share this? 

Would basically be a global/constant, e.g.:

```ruby
module Sqlite3
  FORK_LOCK = Concurrent::ReadWriteLock .new
end
```

> What operations would be behind the read lock?

Anything requiring the connection not to be closed. So `query`, `transaction do`, etc.

> a global shared/exclusive lock owned by the SQLite3Adapter class

I was thinking of doing this in Sqlite3-ruby, not active record. Because the idea isn't to lock when a connection is checked out, but when it's in active use. This way if a thread is still holding a connection but not actiely using it, we can close + `sqlite3_free` under its nose. And then later when the fork lock is released and a thread try to use such connection again, we recreate it on the fly.

Of course there are probably complications I'm not thinking of, like prepared statements would be invalidated? etc.

flavorjones:
@jeremy suggested changing `Database#close` to call `Database#discard` if the process ID has changed. I really like this idea, and emitting the warning from the gem will help alert people to when they're doing something that sqlite doesn't support. I've added a commit to the PR to do this.

adr/2024-09-fork-safety.md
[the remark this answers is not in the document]
flavorjones: > prepared statements would be invalidated

Yes, this is what I found to be the problem on the branch I linked above.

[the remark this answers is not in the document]
byroot: > I would love to chat with you and other folks (maybe at Rails World?)

Yeah, I'm sure we can find time with Matthew to chat about that. Dunno if there's really a good solution to this problem, but...

byroot:
> suggested changing `Database#close` to call `Database#discard` if the process ID has changed. I really like this idea, and emitting the warning from the gem will help alert people to when they're doing something that sqlite doesn't support.

Agreed. That's a great idea. Make it less weird than exposing a method that basically says: leak memory for me please.

flavorjones:
I've rewritten this PR. Diffs from the original:

- there is no `Database#discard` method being introduced
- instead, any time a database is closed (either explicitly via `Database#close` or implicitly via GC) the current pid is checked against the creating pid and if they are different the connection is "discarded" and a warning is emitted

After conversations with a few folks, I came to the conclusion that a `discard` method was just confusing and unnecessary.

I'll leave this open for a bit if anyone wants to give feedback.

ext/sqlite3/database.c
```diff
@@ -12,6 +12,39 @@
 
 VALUE cSqlite3Database;
 
+static void
+close_or_discard_db(sqlite3RubyPtr ctx)
+{
+    if (ctx->db) {
+        if (ctx->owner == getpid()) {
```
byroot, line 19: NB: on recent `glibc`, `getpid` is no longer cached, so this is an extra syscall. 

But I assume `allocate/close` are not supposed to be hot spots, so it's probably fine.

ext/sqlite3/database.c
```diff
@@ -12,6 +12,39 @@
...
+            // "Discard" it. See adr/2024-09-fork-safety.md
+            sqlite3_file *sfile;
+            int status;
+
+            rb_warning("An open sqlite database connection was inherited from a forked process and "
+                       "is being discarded. This is a memory leak. If possible, please close all sqlite "
+                       "database connections before forking.");
+
+            // close the open file descriptors
+            status = sqlite3_file_control(ctx->db, NULL, SQLITE_FCNTL_FILE_POINTER, &sfile);
```
byroot, line 33: `dbName` is always `NULL`? I can't find in the API in which case you have other databases open on a single connection.

ext/sqlite3/database.c
```diff
@@ -12,6 +12,39 @@
...
+        if (ctx->owner == getpid()) {
+            // Ordinary close.
+            sqlite3_close_v2(ctx->db);
+        } else {
+            // This is an open connection carried across a fork().
+            // "Discard" it. See adr/2024-09-fork-safety.md
+            sqlite3_file *sfile;
+            int status;
+
+            rb_warning("An open sqlite database connection was inherited from a forked process and "
```
byroot, line 28: I'm starting to wonder? This doesn't apply to read only connections right? Might be nice to not leak for them?

ext/sqlite3/database.c
flavorjones: Good to know! I am fully OK with this.

ext/sqlite3/database.c
flavorjones: From https://sqlite.org/c3ref/file_control.html :

> The name of the database is "main" for the main database or "temp" for the TEMP database, or the name that appears after the AS keyword for databases that are added using the [ATTACH](https://sqlite.org/lang_attach.html) SQL command. A NULL pointer can be used in place of "main" to refer to the main database file.

byroot: Right so potentially you'd need to close more than just the main database?

ext/sqlite3/database.c
flavorjones: > This doesn't apply to read only connections right

My reading of §2.6 of https://www.sqlite.org/howtocorrupt.html does not exclude readonly connections. I see it mentions "locking problems" but I don't think that sentence is intended to be a canonical list of conditions. I'm choosing to pay attention to:

> SQLite is not designed to support that kind of behavior. Any database connection that is used in a child process must be opened in the child process.

which is unconditional.

I'm choosing to be extremely conservative here because I think the negative impact of data loss justifies it. We can always loosen the "discard" conditions if we learn more.

byroot: > I'm choosing to be extremely conservative here because I think the negative impact of data loss justifies it.

Yeah, but if the connection is read only, I don't see how there could possibly be data loss. I have some planned used for read only sqlite as some static data store, and not leaking memory on every refork would be preferable.

flavorjones:
@tenderlove Can I get your :eyes: on this before I whack merge

ext/sqlite3/database.c
flavorjones: Keep in mind I don't fully understand the nature of how sqlite implements locking, but reading through this entry in particular seems relevant:

> suppose a multi-thread process has two or more threads with separate SQLite database connections to the same database file. Then a third thread comes along and wants to read something out of that same database file on its own, without using the SQLite library. The third thread does an open(), a read() and then a close(). One would think this would be harmless. But the close() system call caused the locks held on the database by all the other threads to be dropped. Those other threads have no way of knowing that their locks have just been trashed

The implication here is that another thread in the process could be writing, and the presence of a read-only connection is still going to interact with locks on the database and could screw up the writer(s).

I suppose we could implement tracking of which `Database` objects are pointed at the same database file and which are writers and have additional logic in the `close_or_discard_db` function ... but that doesn't feel justified ... 🤷 why not just open the readonly connection in the child?

ext/sqlite3/database.c
flavorjones: Ah, I misunderstood what you were saying. I'll poke around.

flavorjones: Relevant section of the docs is https://www.sqlite.org/inmemorydb.html

ext/sqlite3/database.c
flavorjones: Like, I'm open to adding more specific conditions to the check we're doing ... but not in this PR. I want to land this and cut a release. Then we can take the time to make it work more optimally for some of these cases. Does that sound fair?

byroot: Yeah, of course.

tenderlove: > Like, I'm open to adding more specific conditions to the check we're doing ... but not in this PR.

👍 

Since you _can_ open a database read-only, it seems fine to do more specific checks. I agree with shipping this first though.

tenderlove (approved):
This makes sense, but if I understand it correctly the change doesn't prevent corruption, it just emits a warning if someone calls "close" on a database that happened to be opened in another process.  Someone could, for example, fork a process then execute queries in the child process thereby causing corruption.

I don't think we should check the PID for every query.  Unsure if it's possible, but could we register a callback with `Process._fork` for every open database and then issue a warning on fork?

To be clear, I'm fine with the change in this PR.  It makes sense.

ext/sqlite3/database.c
flavorjones: Just pushed a commit that should make sure we get all the named databases.

flavorjones:
> if I understand it correctly the change doesn't prevent corruption

It does prevent corruption! By not calling `sqlite3_close_v2` we avoid corruption in the case demonstrated in https://github.com/flavorjones/2024-09-13-sqlite-corruption.

flavorjones:
> could we register a callback with Process._fork for every open database and then issue a warning on fork?

Yes, we could absolutely do that -- just to make sure I understand, you want the warning issued in the parent process when the fork happens, yes? I may push it to a separate PR though.

flavorjones:
@tenderlove Thanks for reviewing.

> if I understand it correctly the change doesn't prevent corruption

It will prevent corruption for Rails apps (when we also update sqlite3adapter). By not calling `sqlite3_close_v2` we avoid corruption in the case demonstrated in https://github.com/flavorjones/2024-09-13-sqlite-corruption.

> Someone could, for example, fork a process then execute queries in the child process thereby causing corruption

but yes, it doesn't _always_ prevent corruption, I'm just trying to nail the Rails use case here (which will require a PR to Rails as well).

> could we register a callback with Process._fork for every open database and then issue a warning on fork?

Yes, we could absolutely do that -- just to make sure I understand, you want the warning issued in the parent process when the fork happens, yes? I may push it to a separate PR though.

I guess we could also register a callback with Process._fork that would immediately close ("discard") the connection in the child. That might prevent corruption anywhere. Let me play with that.

flavorjones:
I've added another commit that will call `sqlite3_db_release_memory` to free cache pages, etc. which should buy us something on heavily-used connections.

flavorjones:
I've pushed one more commit that does some big changes that I'd like a re-review of.

- Readonly connections are excepted from all special `close` handling.
- The gem hooks into `Process._fork` and automatically closes all open, writable connections after fork.

@tenderlove @byroot Thanks for your feedback so far.

ext/sqlite3/database.c
```diff
@@ -12,6 +12,59 @@
...
+        } else {
+            // This is an open connection carried across a fork().
+            // "Discard" it. See adr/2024-09-fork-safety.md
+            sqlite3_file *sfile;
+            int status;
+
+            rb_warning("An open writable sqlite database connection was inherited from a "
+                       "forked process and is being closed to prevent the risk of corruption. "
+                       "If possible, please close all writable sqlite database connections "
+                       "before forking.");
```
tenderlove, line 33: Do we still need this warning if we're auto-closing stuff cross fork?  I'm OK with keeping the warning. IMO if you leave the connection open and fork then it's probably a bug in your app and we should let the end user know.

flavorjones: Yeah, thanks for asking, this is a good question. I am worried that this change is going to break somebody's app ("Why is this database connection closed?") and they'll have to dig around to figure out what's going on. But I guess that's what release notes are for? I could go either way but tend to err towards overcommunication.

tenderlove: > I could go either way but tend to err towards overcommunication.

💯

flavorjones: If people get annoyed we can remove it later.

lib/sqlite3/fork_safety.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    @databases = []
+
+    class << self
+      def hook!
+        ::Process.singleton_class.prepend(CoreExt)
+      end
+
+      def track(database)
+        @databases << WeakRef.new(database)
```
flavorjones, line 26: Do we need to make this thread-safe?

flavorjones: Or should I just use `ObjectSpace.each_object(Database)`?

lib/sqlite3/fork_safety.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require "weakref"
+
+# based on Rails's active_support/fork_tracker.rb
+module SQLite3
+  module ForkSafety
+    module CoreExt
+      def _fork
```
flavorjones, line 9: This is only going to work in Ruby 3.1+ so we'll need to remember to drop support for Ruby 3.0 (which it's time to do, anyway).

flavorjones: Done in an additional commit.

lib/sqlite3/fork_safety.rb
flavorjones: Made it thread-safe.

tenderlove: > Do we need to make this thread-safe?

🤷🏻‍♀️

If we're really worried about it, we could use a Queue object. Then we don't need to pull in another library.

tenderlove: > Or should I just use `ObjectSpace.each_object(Database)` instead of keeping our own list?

I like the list better than iterating the heap.

flavorjones:
Added a commit that ensures Statements related to a discarded database will raise an exception if people try to use them.

flavorjones:
Will cut a release candidate today.

flavorjones:
https://github.com/sparklemotion/sqlite3-ruby/releases/tag/v2.1.0.rc1

ext/sqlite3/statement.c
```diff
@@ -121,6 +135,7 @@ step(VALUE self)
 
     TypedData_Get_Struct(self, sqlite3StmtRuby, &statement_type, ctx);
 
+    require_open_db(self);
```
tenderlove, line 138: I didn't notice this in the initial review. `step` is an extremely hot method, do we actually need to do this?  I thought SQLite would already raise an exception if the db is somehow closed out from under it.

byroot: Yeah, it does an IV read which is a bit much. If it was a struct member access I think it would be acceptable.

flavorjones: I will look at other ways to make sure we're protected. This is really just protecting us from default if the DB was "discarded" so I can set a bit in the struct and that should be good enough.

flavorjones: See https://github.com/sparklemotion/sqlite3-ruby/pull/565

ext/sqlite3/database.c
flavorjones: See https://github.com/sparklemotion/sqlite3-ruby/pull/566

PR #565, opened 2024-09-19
Optimize the statement check for a non-discarded database

### LINKED ISSUE #564: Use `sqlite3_close` instead of `sqlite3_close_v2`?
I didn't fully realize the consequences of switching to `sqlite3_close_v2` when I reviewed #558.  Since `sqlite3_close_v2` doesn't return a status code, it means that we can have `SQLite3::Statement` objects that point at a database which has been closed by someone.

Since the database could have been closed, we have to check on every call to `step` whether or not the db is "ok", and unfortunately the `step` method is extremely hot. It's called for every row of the statement.

For example, this benchmark:

```ruby
require 'sqlite3'
require 'benchmark/ips'

file = "/tmp/test.sqlite3"

File.unlink file rescue nil

db = SQLite3::Database.new file
db.execute "create table foo (id int)"

1000.times do |i|
  db.execute "insert into foo (id) values (#{i})"
end

Benchmark.ips do |x|
  x.report("selecting") {
    db.execute("select * from foo").to_a
  }
end

db.close
```

At commit 480f3e7d, the results are like this:

```
$ ruby -I lib test.rb
ruby 3.3.4 (2024-07-16 revision 425e468d25) [arm64-darwin23]
Warming up --------------------------------------
           selecting   855.000 i/100ms
Calculating -------------------------------------
           selecting      8.495k (± 0.4%) i/s -     42.750k in   5.032394s
```

At commit 81ea485f680c54f604ec1d0f40a00d999a149dc0, the results are like this:

```
$ ruby -I lib test.rb
ruby 3.3.4 (2024-07-16 revision 425e468d25) [arm64-darwin23]
Warming up --------------------------------------
           selecting   582.000 i/100ms
Calculating -------------------------------------
           selecting      5.810k (± 0.5%) i/s -     29.100k in   5.008952s
```

This simple benchmark is substantially slower after switching to `sqlite3_close_v2`.

Before da908657099a20f2796e49e6ed42c102a9aa7a80, the following code would raise an exception:

```ruby
require 'sqlite3'
require 'benchmark/ips'

file = "/tmp/test.sqlite3"

File.unlink file rescue nil

db = SQLite3::Database.new file
db.execute "create table foo (id int)"

1000.times do |i|
  db.execute "insert into foo (id) values (#{i})"
end

stmt = db.prepare "select * from foo"
stmt.step

db.close # used to raise an exception
```

This meant that statements couldn't have "dangling references" to a database connection.  Since this doesn't raise an exception anymore, it is possible and we must therefore check the validity of the db connection before fetching the next row.

Would it be possible to switch back to using `sqlite3_close` and raising an exception?

I know the SQLite documentation [says we should call `sqlite3_close_v2` in GC languages](https://www.sqlite.org/c3ref/close.html), but I think it only makes sense to call that function from the free function called by the GC.  If a _user_ calls `close`, we should use `sqlite3_close` and raise an exception.

Statements [keep a reference to the database connection](https://github.com/sparklemotion/sqlite3-ruby/blob/81ea485f680c54f604ec1d0f40a00d999a149dc0/lib/sqlite3/statement.rb#L33), so we shouldn't have to fear the db connection being closed via GC until the user is done using the statement.  I've read #558 a few times, and it's not clear to me why `sqlite3_close_v2` is necessary for fork safety, but I could definitely be missing something.

Thanks.

flavorjones:
Closes #564 which pointed out a performance regression from #558.

This also restores the ability (now tested!) to call `Database#close` successfully and defer its cleanup until after `Statement`s are closed, which was the promise of #557 and `sqlite3_close_v2`.

Running the benchmark @tenderlove wrote for #564:

```ruby
#!/usr/bin/env ruby

require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "sqlite3", path: "."
  gem "benchmark-ips"
end

require "sqlite3"
require "benchmark/ips"

file = "/tmp/test.sqlite3"

begin
  File.unlink file
rescue
  nil
end

db = SQLite3::Database.new file
db.execute "create table foo (id int)"

1000.times do |i|
  db.execute "insert into foo (id) values (#{i})"
end

Benchmark.ips do |x|
  x.report("selecting") {
    db.execute("select * from foo").to_a
  }
end

db.close
```

On 480f3e7 after `sqlite3_open_v2` was introduced by #557:

```
ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [x86_64-linux]
Warming up --------------------------------------
           selecting   533.000 i/100ms
Calculating -------------------------------------
           selecting      5.306k (± 1.3%) i/s  (188.47 μs/i) -     26.650k in   5.023715s
```

On 81ea485f after fork protection was introduced by #558, performance regressed:

```
ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [x86_64-linux]
Warming up --------------------------------------
           selecting   329.000 i/100ms
Calculating -------------------------------------
           selecting      3.261k (± 3.4%) i/s  (306.66 μs/i) -     16.450k in   5.051880s
```

On 56d47a67 in this PR, performance is restored.

```
ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [x86_64-linux]
Warming up --------------------------------------
           selecting   548.000 i/100ms
Calculating -------------------------------------
           selecting      5.476k (± 1.7%) i/s  (182.61 μs/i) -     27.400k in   5.004987s
```

### DIFF
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -49,15 +49,16 @@ discard_db(sqlite3RubyPtr ctx)
     }
 
     ctx->db = NULL;
+    ctx->flags |= SQLITE3_RB_DATABASE_DISCARDED;
 }
 
 static void
 close_or_discard_db(sqlite3RubyPtr ctx)
 {
     if (ctx->db) {
-        int isReadonly = (ctx->flags & SQLITE_OPEN_READONLY);
+        int is_readonly = (ctx->flags & SQLITE3_RB_DATABASE_READONLY);
 
-        if (isReadonly || ctx->owner == getpid()) {
+        if (is_readonly || ctx->owner == getpid()) {
             // Ordinary close.
             sqlite3_close_v2(ctx->db);
             ctx->db = NULL;
@@ -153,7 +154,9 @@ rb_sqlite3_open_v2(VALUE self, VALUE file, VALUE mode, VALUE zvfs)
              );
 
     CHECK(ctx->db, status);
-    ctx->flags = flags;
+    if (flags & SQLITE_OPEN_READONLY) {
+        ctx->flags |= SQLITE3_RB_DATABASE_READONLY;
+    }
 
     return self;
 }
@@ -943,11 +946,10 @@ rb_sqlite3_open16(VALUE self, VALUE file)
 #endif
 #endif
 
-    status = sqlite3_open16(utf16_string_value_ptr(file), &ctx->db);
-
-    // these are the perm flags used implicitly by sqlite3_open16,
+    // sqlite3_open16 implicitly uses flags (SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE)
     // see https://www.sqlite.org/capi3ref.html#sqlite3_open
-    ctx->flags = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE;
+    // so we do not ever set SQLITE3_RB_DATABASE_READONLY in ctx->flags
+    status = sqlite3_open16(utf16_string_value_ptr(file), &ctx->db);
 
     CHECK(ctx->db, status)
 
--- a/ext/sqlite3/database.h
+++ b/ext/sqlite3/database.h
@@ -3,6 +3,10 @@
 
 #include <sqlite3_ruby.h>
 
+/* bits in the `flags` field */
+#define SQLITE3_RB_DATABASE_READONLY  0x01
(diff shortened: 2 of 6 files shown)


flavorjones:
@tenderlove I'd like to merge this and cut an RC2 today, but would love your eyes on this if you get a chance.

tenderlove:
@flavorjones sorry was busy today. This is great though, thank you for fixing it!

PR #571, opened 2024-10-28
feat: allow URI filenames

flavorjones:
See https://www.sqlite.org/uri.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## next / unreleased
 
+### Added
+
+- URI filenames are now allowed. This allows the injection of some behavior via recognized query parameters. See https://www.sqlite.org/uri.html for more information. #571 @flavorjones
+
+
 ### Improved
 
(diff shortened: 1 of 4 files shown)


flavorjones:
I think it might be better to not merge this -- the `Open::URI` bit can be passed where people want to pass a URI, and all the options available via URI are also available via the API.

What I really want is to be able to inject configuration options in a Rails app via the `config/database.yml` and so maybe it's better for me to do some work in the Rails sqlite3 adapter instead.

flavorjones:
Yeah, going to move this decision to Rails.

flavorjones:
Hrm, reopening.

flavorjones:
FWIW, it looks like the major distros compile with `-DSQLITE_USE_URI=1`, also (e.g. debian has compiled with it since 2019 / 3.27.1-1).

tenderlove:
Ah, I didn't realize this was a feature natively supported by SQLite.  Yes, we should do this.

Issue #559, opened 2024-09-16
Incomplete PRAGMA support


rickhull:
I have noticed some inconsistencies in the handling for SQLite pragmas.

1. There is a setter for `busy_timeout=` but no getter.  Likewise for `ignore_check_constraints=`.
2. No accessors for `analysis_limit` `hard_heap_limit` `trusted_schema`
3. No methods for `function_list` `module_list` `pragma_list`
4. No methods for `table_list` `table_xinfo`

SQLite supports 3 pragmas for index and table:

* thing_list
* thing_info
* thing_xinfo

The gem I have installed, **version 2.0.4**, has methods for all three index methods but only the `table_info` method.

Some of these may not make sense to support, but these are my findings.  Separately, I have some ideas about how to reorganize lib/sqlite3/pragmas.rb.

rickhull:
I might take a stab.  Here's what I'm doing in the meantime: https://gist.github.com/rickhull/464c6fdf5fdea217894f67618b4a2b0c

rickhull:
Please take a look at the `Nostrb::SQLite::Pragma` class: https://github.com/rickhull/nostrb/blob/master/lib/nostrb/sqlite.rb#L6

I want to basically import this wholesale, or something like it, as our pragma backend.  We can then patch or rewrite the current API using this backend.  I think it will be relatively easy to delete nearly of the current pragma impl with this more programmatic approach.

Even if not, I may still look at extending the current approach.

flavorjones:
@rickhull I'm not opposed to doing something like what you're describing, but maybe we can get there iteratively instead of big-bang copying code and then trying to adapt to it.

For example, I really like how that class defines the pragma setters and getters dynamically. Maybe we could start by introducing that technique into this gem (and adding some tests for those pragmas)? I think that really cleans things up and addresses almost all of the initial issues you raised above. WDYT?

rickhull:
Totally agreed on not necessarily big-bang-adapt.  If I can land a suitable patch here, then I can delete most of my stuff.  Let me come up with something more concrete in the next few days.

rickhull:
How do we feel about these pragma I've ignored?

```
      #      debug: parser_trace schema_version stats vdbe_* writable_schema
      #     legacy: legacy_alter_table legacy_file_format
      # deprecated: case_sensitive_like count_changes data_store_directory
      #             default_cache_size empty_result_callbacks full_column_names
      #             short_column_names temp_store_directory
```

[...]

PR #572, opened 2024-10-29
Add convenience wrapper for pragma optimize

alexcwatt:
The SQLite docs recommend some cases where `PRAGMA optimize;` should be used: https://www.sqlite.org/pragma.html#pragma_optimize

I have discovered recently a case where this helped a project at work (https://twitter.com/alexcwatt/status/1851094243323912631) and now I'm wondering about how to make this easier to discover and use.

Do we want a way to test that these wrappers translate to the expected `execute` call? I wasn't sure what tests to write for this change.

### DIFF
--- a/lib/sqlite3/pragmas.rb
+++ b/lib/sqlite3/pragmas.rb
@@ -338,6 +338,14 @@ module SQLite3
       set_int_pragma "mmap_size", size
     end
 
+    def optimize(bitmask = nil)
+      if bitmask
+        set_int_pragma "optimize", bitmask
+      else
+        execute("PRAGMA optimize")
+      end
+    end
+
     def page_count
       get_int_pragma "page_count"
     end
--- a/lib/sqlite3/pragmas.rb
+++ b/lib/sqlite3/pragmas.rb
@@ -338,6 +338,10 @@ module SQLite3
       set_int_pragma "mmap_size", size
     end
 
+    # Attempt to optimize the database.
+    #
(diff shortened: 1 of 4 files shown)


flavorjones:
@alexcwatt Thanks for this! The omission was mentioned by @rickhull in https://github.com/sparklemotion/sqlite3-ruby/issues/559

If we're going to allow the optimize bits to be passed in by the caller, I'd like the API to be a bit more developer-friendly. Would you consider defining some constants in `lib/sqlite3/constants.rb`?

Our test coverage is pretty bad right now for the pragma methods, so as long as we keep the implementation simple, I think it's probably OK to not have test coverage.

lib/sqlite3/pragmas.rb
```diff
@@ -338,6 +338,14 @@ def mmap_size=(size)
       set_int_pragma "mmap_size", size
     end
 
+    def optimize(bitmask = nil)
+      if bitmask
```
flavorjones, line 342: I'd like to see these bits made available as constants like many of the other sqlite bitmasks in `lib/sqlite3/constants.rb`

flavorjones (changes requested):

lib/sqlite3/pragmas.rb
alexcwatt: That's a good idea. I took a pass at this; please let me know what you think.

flavorjones (approved):
Looks good! I added a `DEFAULT` constant and some tests.

PR #578, opened 2024-11-12
Compile support for SQLITE_DBPAGE

flavorjones:
https://www.sqlite.org/dbpage.html

cc @kevinmcconnell


kevinmcconnell:
Nice! 🎉

flavorjones:
I may need to introduce a constant to track whether the libsqlite being used is vendored or a system library. It's useful to know when to skip the sqlite_dbpage test, but is also generally useful for the gem to be more self-descriptive for debugging purposes (e.g., `Nokogiri::VERSION_INFO`).

Issue #573, opened 2024-10-29
Evaluate whether to adopt any compiler flags set by debian


flavorjones:
Debian currently sets these flags (as of sid/unstable 3.46.1-1):

```
export CFLAGS += -O2 -fno-strict-aliasing \
    -DSQLITE_SECURE_DELETE -DSQLITE_ENABLE_COLUMN_METADATA \
    -DSQLITE_ENABLE_FTS3 -DSQLITE_ENABLE_FTS3_PARENTHESIS \
    -DSQLITE_ENABLE_RTREE=1 -DSQLITE_SOUNDEX=1 \
    -DSQLITE_ENABLE_UNLOCK_NOTIFY \
    -DSQLITE_ENABLE_DBSTAT_VTAB \
    -DSQLITE_ALLOW_ROWID_IN_VIEW \
    -DSQLITE_ENABLE_UPDATE_DELETE_LIMIT=1 \
    -DSQLITE_ENABLE_LOAD_EXTENSION \
    -DSQLITE_ENABLE_JSON1 \
    -DSQLITE_LIKE_DOESNT_MATCH_BLOBS \
    -DSQLITE_THREADSAFE=1 \
    -DSQLITE_ENABLE_FTS3_TOKENIZER=1 \
    -DSQLITE_USE_URI=1 \
    -DSQLITE_MAX_SCHEMA_RETRY=25 \
    -DSQLITE_ENABLE_PREUPDATE_HOOK \
    -DSQLITE_ENABLE_SESSION \
    -DSQLITE_ENABLE_STMTVTAB \
    -DSQLITE_STRICT_SUBTYPE=1 \
    -DSQLITE_MAX_VARIABLE_NUMBER=250000
```

Any of these make sense for this gem to set when it compiles?

flavorjones:
Because https://github.com/sparklemotion/sqlite3-ruby/pull/605 is showing that the new autoconf amalgamations are NOT working properly, I'm working on using the plain amalgamation and will pull these preprocessor defs into that PR.

flavorjones:
Related, see https://github.com/sparklemotion/sqlite3-ruby/pull/636 which is a branch using the plain amalgamation and these CPP flags.

PR #580, opened 2024-11-18
Enable dbstat table by default

pawurb:
Hi. Similar to https://github.com/TryGhost/node-sqlite3/issues/1279 . This PR enables `dbstat` table, that contains useful info like size of tables, indexes, etc. useful for debugging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - The SQLITE_DBPAGE extension is now enabled by default, which implements an eponymous-only virtual table that provides direct access to the underlying database file by interacting with the pager. See https://www.sqlite.org/dbpage.html for more information. [#578] @flavorjones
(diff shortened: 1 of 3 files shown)


flavorjones:
Nice! Thank you for opening this.

Documentation on `DBSTAT` is here: https://sqlite.org/dbstat.html

Related to https://github.com/sparklemotion/sqlite3-ruby/issues/573 which documents Debian has this flag set already.

@pawurb Can you please add a small test like the one in https://github.com/sparklemotion/sqlite3-ruby/pull/578/files to exercise the virtual table and prevent regression? The test will may fail on some platforms, in which case I'll need the same sort of metadata ("is this using the vendored sqlite or a system sqlite?") to conditionally run the test (and I can add that if needed).

flavorjones:
Rebased and added a commit to skip the test when using system libraries (specifically, if the user links against sqlcipher or mswin/mingw sqlite -- see https://github.com/sparklemotion/sqlite3-ruby/actions/runs/11910233460 for the failing tests).

pawurb:
thanks for a quick merge! I've started building https://github.com/pawurb/rails-sqlite-extras so it will help a lot

flavorjones:
😍 that project looks neat! Hoping to ship a release of this today.

PR #590, opened 2024-11-25
ci: remove gem-update-system

flavorjones:
which may not be necessary anymore? It's slow on the aarch64-linux systems so let's try to remove it.


flavorjones:
Oh man, CI time dropped from ~25 minutes to ~5 minutes. Nice.

PR #591, opened 2024-11-25
test: clean up unnecessary ensure/cleanup blocks

flavorjones:
I'm pretty sure it's unnecessary to explicitly close databases and statements post-0c24631d and related changes.


flavorjones:
If we don't explicitly close during teardown, we'll get warnings from the after-fork tests.

PR #592, opened 2024-12-04
fix: handle the case where GC runs during fork safety checks

flavorjones:
If GC runs between the check for `db.weakref_alive?` and `db.close` then an exception will be raised:

    Invalid Reference - probably recycled (WeakRef::RefError)

I personally saw this error when running a Rails app test suite in parallel, so while it's rare it does happen.

In this case, let's just swallow the error and keep going, since the database object isn't in use and was GCed.

### DIFF
--- a/lib/sqlite3/fork_safety.rb
+++ b/lib/sqlite3/fork_safety.rb
@@ -35,17 +35,21 @@ module SQLite3
         @databases.each do |db|
           next unless db.weakref_alive?
 
-          unless db.closed? || db.readonly?
-            unless warned
-              # If you are here, you may want to read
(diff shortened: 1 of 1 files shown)


flavorjones:
@tenderlove or @byroot can I ask for a quick review of this?

I considered using ObjectSpace::WeakMap because its foreach has a GC guard on both key and value; but it doesn't have a `#clear` method so it didn't feel any cleaner than this PR.

byroot:
Yeah, the missing `#clear` is a bit annoying but you can always replace the map entirely (but doesn't allow to store it in constant).

byroot (approved):
Looks good to me.

PR #605, opened 2025-01-14
dep: update vendored sqlite to 3.49.1

flavorjones:
This skips 3.48.0 and 3.49.0 because the autoconf package is not building properly across all platforms, as noted in comments below.

- https://sqlite.org/releaselog/3_48_0.html
- https://sqlite.org/releaselog/3_49_0.html
- https://sqlite.org/releaselog/3_49_1.html

Note that this PR also updates tests to not expect to be able to load extensions when running on Windows systems using the packaged library. The autoconf amalgamation is not picking up dlopen during configuration on Windows systems and I have been unable to figure out why.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Dependencies
 
-- Vendored sqlite is updated to [v3.49.0](https://sqlite.org/releaselog/3_49.0.html) (from v3.47.2). #605 @flavorjones
+- Vendored sqlite is updated to [v3.49.1](https://sqlite.org/releaselog/3_49_1.html) (from v3.47.2). #605 @flavorjones
 - Updated to rake-compiler-dock v1.9.1. #610 @flavorjones
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Dependencies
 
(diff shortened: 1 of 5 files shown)


flavorjones:
This is failing in CI because the pkgconf file being generated by 3.48.0 is not correct. I opened a thread upstream to discuss:

https://sqlite.org/forum/forumpost/35c2397b15

flavorjones:
Jumping to 3.49.0.

flavorjones:
The build is breaking on windows now, sigh.

I'm going to try to use the single-file amalgamation because this autoconf change in the upstream is pretty obviously not being used wide in the autoconf amalgamation file (it's been broken in different ways for two major releases now). Will take me a few days to circle back.

flavorjones:
Note that this PR also updates tests to not expect to be able to load extensions when running on Windows systems using the packaged library. The autoconf amalgamation is not picking up dlopen during configuration on Windows systems, and I have been unable to figure out why.

Upstream conversation happened at https://sqlite.org/forum/forumpost/dc98a0daa496cf929fe95c5fb62a2305cca3c7e8c20d20903bcb813b9082e390. But I've timed out on trying to figure out how to fix this. If there are windows users out there who need/want/use this feature, I would love your help fixing/reporting it.

PR #620, opened 2025-03-06
Freeze strings and constants in pragmas

tenderlove:
I'm looking in earnest at Ractor support in Rails. We need some of these constants that SQLite3 uses to be frozen so that when they're accessed via Ractors we don't get an error.

This commit just adds frozen string literals, and freezes some constants

### DIFF
--- a/lib/sqlite3/pragmas.rb
+++ b/lib/sqlite3/pragmas.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require "sqlite3/errors"
 
 module SQLite3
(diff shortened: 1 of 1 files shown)


flavorjones:
@tenderlove LMK if you need this in a release.

flavorjones:
Side note: the freezing idiom here is kinda gross. Is there no better way to do this, like `Ractor.make_shareable` or `# shareable_constant_value: literal`?

I read through [Feature #17145: Ractor-aware `Object#deep_freeze`](https://bugs.ruby-lang.org/issues/17145) and [Feature #17274: Ractor.make_shareable(obj)](https://bugs.ruby-lang.org/issues/17274) and I also skimmed [ruby/doc/ractor.md](https://github.com/ruby/ruby/blob/master/doc/ractor.md) I'm still not sure I understand how this is _supposed_ to be done.

tenderlove:
> @tenderlove LMK if you need this in a release.

I do not.  I'm working off edge here, so it's fine.



> Side note: the freezing idiom here is kinda gross. Is there no better way to do this, like `Ractor.make_shareable` or `# shareable_constant_value: literal`?

I didn't want to use the `Ractor` constant because then the code isn't backwards compatible (or we'd have to do work to make it backwards compatible). I think the comment directive might work, but IIRC there's a problem with it (that I can't remember off the top of my head). I just recall an issue while fixing a bug in Prism (not an issue with Prism, but a problem with how the directive works in practice).

I agree we're calling `freeze` a bunch.  Maybe we should try to figure out why the data structure is nested so much?  It looks like some of these constants are truly tuples, so maybe we could use a hash.  I'm not sure why the other ones are nested arrays.  I'll look in to it.

PR #621, opened 2025-03-07
Try to simplify pragma related constants

tenderlove:
This PR is just trying to simplify pragma related constants.  I guess these constants are used as a kind of lookup table.  I added some regression tests because I saw we're not exercising all cases in the test suite, then refactored the constants to just be hash tables.

I left the very strange array search in `set_enum_pragma` for backwards compatibility, though I think we should deprecate it.

AFAICT, there are two types of pragmas. One uses integers, the other strings.  I think if we changed some of the pragma APIs, we could delete the hash tables that are simply duplicated strings, but I'll do that in a follow up PR.

### DIFF
--- a/lib/sqlite3/database.rb
+++ b/lib/sqlite3/database.rb
@@ -196,7 +196,7 @@ module SQLite3
     #
     # Fetch the encoding set on this database
     def encoding
-      prepare("PRAGMA encoding") { |stmt| Encoding.find(stmt.first.first) }
+      Encoding.find super
     end
 
     # Installs (or removes) a block that will be invoked for every access
--- a/lib/sqlite3/pragmas.rb
+++ b/lib/sqlite3/pragmas.rb
@@ -269,7 +269,7 @@ module SQLite3
     end
 
     def encoding=(mode)
-      set_enum_pragma "encoding", mode, ENCODINGS
+      set_string_pragma "encoding", mode, ENCODINGS
     end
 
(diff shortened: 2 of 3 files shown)


flavorjones:
@tenderlove Looks like there's a related test failure in the downstream activerecord test suite, I haven't dug into it yet and probably won't get a chance today.

tenderlove:
> @tenderlove Looks like there's a related test failure in the downstream activerecord test suite, I haven't dug into it yet and probably won't get a chance today.

I tracked it down to [this](https://github.com/rails/rails/blob/c218fb6cd2bdd67771aac5bfce8097e43c05a26c/activerecord/test/cases/adapters/sqlite3/sqlite3_adapter_test.rb#L246-L251), and I think I've fixed it in the latest commit.

Rails was testing that if it passes an invalid pragma name, we get an exception.  It seems like our code (the SQLite3 Ruby bindings) are the things validating what pragmas are valid.  I don't like that we're in the business of validating these because if SQLite3 ever changes then we also have to update our code. Not to mention, some pragmas might be valid on some versions of SQLite3 and not on others.  I feel we should take a "garbage in garbage out" approach here, though that might mean a major version bump.  I've made the code backwards compatible for now though.

flavorjones:
> I think I've fixed it in the latest commit

Yup, it's green! Whacking auto-merge.

> I don't like that we're in the business of validating these

Agree.

Issue #634, opened 2025-07-03
gem install sqlite3 --platform=ruby fails to build starting at 2.6.0


copiousfreetime:
## Installing the ruby platform gem

Admittedly I'm doing this on an older linux -- rockylinux 8.6 -- which is linux kernel `5.19.11-1ep.el8.x86_64` and glibc 2.28.

But that shouldn't affect a working ruby 3.4.4, from building sqlite3.

It does look like this started with sqlite3 2.6.0 which is also when the autotools was changed to autosetup -- so I wonder if that is actually the issue.

## Gem Environment
```
RubyGems Environment:
  - RUBYGEMS VERSION: 3.6.9
  - RUBY VERSION: 3.4.4 (2025-05-14 patchlevel 34) [x86_64-linux]
  - INSTALLATION DIRECTORY: /home/vagrant/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0
  - USER INSTALLATION DIRECTORY: /home/vagrant/.gem/ruby/3.4.0
  - CREDENTIALS FILE: /home/vagrant/.local/share/gem/credentials
  - RUBY EXECUTABLE: /home/vagrant/.rbenv/versions/3.4.4/bin/ruby
  - GIT EXECUTABLE: /usr/bin/git
  - EXECUTABLE DIRECTORY: /home/vagrant/.rbenv/versions/3.4.4/bin
  - SPEC CACHE DIRECTORY: /home/vagrant/.gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /home/vagrant/.rbenv/versions/3.4.4/etc
  - RUBYGEMS PLATFORMS:
     - ruby
     - x86_64-linux
  - GEM PATHS:
     - /home/vagrant/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0
     - /home/vagrant/.gem/ruby/3.4.0
  - GEM CONFIGURATION:
     - :update_sources => true
     - :verbose => true
     - :backtrace => false
     - :bulk_threshold => 1000
     - :sources => ["https://rubygems.org"]
     - :concurrent_downloads => 8
     - "benchmark" => false
     - "gem" => "--no-ri --no-rdoc"
  - REMOTE SOURCES:
     - https://rubygems.org
```

## gem intall sqlite3 --platformr=ruby -v 2.6.0

- [mkmf.log as a gist](https://gist.github.com/copiousfreetime/1f521b7727eb0abdd1a9a67861734970)

[...]

PR #635, opened 2025-07-04
Ensure we can resolve dlopen et al

flavorjones:
Apply the upstream fix from https://github.com/sqlite/sqlite/commit/d9eae6b75a37e63c30c3c05c5c1fc7ed425589cd

Fixes https://github.com/sparklemotion/sqlite3-ruby/issues/634


flavorjones:
Shipped in https://github.com/sparklemotion/sqlite3-ruby/releases/tag/v2.7.2

PR #642, opened 2025-08-22
Implements stmt.named_params

### LINKED ISSUE #627: Expose statement bind parameters
For a given prepared statement, it should be possible to fetch the list of parameters that can be bound to it.

The 3 relevant SQLite methods are:

- https://sqlite.org/c3ref/bind_parameter_count.html
- https://sqlite.org/c3ref/bind_parameter_name.html
- https://sqlite.org/c3ref/bind_parameter_index.html

Without these 3, if you are processing parameters on configurable queries, it is hard to bind _only stuff that's needed_.

For a specific usecase, I do the following in my Jekyll-SQLite extension:

```ruby
##
# Prepare the query by binding the parameters
# Since we don't know if the query needs them
# we ignore all errors about "no such bind parameter"
def _prepare_query(stmt, params)
  params.each do |key, value|
    stmt.bind_param key, value
  rescue StandardError => e
    raise e unless e.message.include? "no such bind parameter"
  end
end
```

params is a dict, but because it is inherited from Jekyll's page data structure, it can contain keys that are not required by the query (such as `title`). So we check the exception and ignore for now.

captn3m0:
Closes #627

Implemented `named_params` directly in C. Not a C programmer, mistakes are mine. Learned about `@` and `$` prefixes while reading the docs, so added them to the test as well.

However, this implementation ignores `sqlite3_bind_parameter_index` entirely, which is only used internally so shouldn't be an issue(?). As of now, the index of the parameter in the result might not match the value returned by `sqlite3_bind_parameter_index` in some cases. 

But this feels better than having to pad the result with nulls.

Once this is finalized, I can update the changelog if needed

### DIFF
--- a/ext/sqlite3/statement.c
+++ b/ext/sqlite3/statement.c
@@ -460,13 +460,13 @@ bind_parameter_count(VALUE self)
     return INT2NUM(sqlite3_bind_parameter_count(ctx->st));
 }
 
-/** call-seq: stmt.params
+/** call-seq: stmt.named_params
  *
- * Return the list of named alphanumeric parameters in the statement.
- * This returns a list of strings.
+ * Return the list of named parameters in the statement.
+ * This returns a frozen array of strings (without the leading prefix character).
  * The values of this list can be used to bind parameters
- * to the statement using bind_param. Numeric and anonymous parameters
- * are ignored.
+ * to the statement using bind_param. Positional (?NNN) and anonymous (?)
+ * parameters are excluded.
  *
  */
 static VALUE
@@ -484,10 +484,9 @@ named_params(VALUE self)
     // The first host parameter has an index of 1, not 0.
     for (int i = 1; i <= param_count; i++) {
         const char *name = sqlite3_bind_parameter_name(ctx->st, i);
-        // We ignore numbered parameters (starting with ?)
-        // And null values, since there can be gaps in the list
+        // We ignore positional and anonymous parameters, and also null values, since there can be
+        // gaps in the list.
         if (name && *name != '?') {
-            // We ignore numeric parameters
             VALUE param = interned_utf8_cstr(name + 1);
             rb_ary_push(params, param);
         }


flavorjones (changes requested):
@captn3m0 Thanks for this! I've kicked off CI.

I'd like to see a test for the behavior of this code when given `?NNN` parameters. Do you think the method should return those parameters? I'm not sure what happens if you try to bind a variable to a name like `?13`.

ext/sqlite3/statement.c
```diff
@@ -460,6 +460,36 @@ bind_parameter_count(VALUE self)
...
+    VALUE params = rb_ary_new2(param_count);
+
+    // The first host parameter has an index of 1, not 0.
+    for (int i = 1; i <= param_count; i++) {
+        const char *name = sqlite3_bind_parameter_name(ctx->st, i);
+        // If parameters of the ?NNN form are used, there may be gaps in the list.
+        if (name) {
+            VALUE rb_name = interned_utf8_cstr(name);
+            // The initial ":" or "$" or "@" or "?" is included as part of the name.
+            rb_name = rb_str_substr(rb_name, 1, RSTRING_LEN(rb_name) - 1);
```
flavorjones, line 486: This creates two String objects which is a little inefficient. How about something like this instead:

```suggestion
            // The initial ":" or "$" or "@" or "?" is included as part of the name.
            VALUE rb_name = interned_utf8_cstr(name + 1);
```

ext/sqlite3/statement.c
```diff
@@ -460,6 +460,36 @@ bind_parameter_count(VALUE self)
...
+    TypedData_Get_Struct(self, sqlite3StmtRuby, &statement_type, ctx);
+
+    REQUIRE_LIVE_DB(ctx);
+    REQUIRE_OPEN_STMT(ctx);
+
+    int param_count = sqlite3_bind_parameter_count(ctx->st);
+    VALUE params = rb_ary_new2(param_count);
+
+    // The first host parameter has an index of 1, not 0.
+    for (int i = 1; i <= param_count; i++) {
```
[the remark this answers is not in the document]
captn3m0: Can add a note about it to the docs.

captn3m0:
Spent a bit more time at the problem, and it looks like the `list bindable numbered params` problem is unsolvable. The `parameter_count` is the max_index, and anything below that is considered bindable. For a statement like `"SELECT ?, ?, ?50"`, you can bind all params from 1-50.

And there's no way to differentiate and get a result that says [1,2,50].

> One way would be to return a mixed array? [2, "foo", "bar", 3], so that the values in the array can be directly used against bind_params. And perhaps change the name merely stmt.params.

I wrote this implementation, which works well as long as params are either numbered or named. Once you add anonymous params to the mix, it goes haywire.

Suggestion: `named_params` should only support _really named params_ (which are alphanumeric tokens), and drop any numeric params. For those, a simple loop from `1-bind_parameter_count` is the best you can do which is already support.

captn3m0:
Pushed a change to ignore numeric and anonymous parameters:

```ruby
stmt = SQLite3::Statement.new(@db, "select  ?1, :foo, ?, $bar, @zed, ?250, @999")
assert_equal ["foo", "bar", "zed"], stmt.named_params
```

captn3m0:
Looked at the windows CI failures, but couldn't figure out why.

test/test_statement.rb
```diff
@@ -256,6 +256,12 @@ def test_named_bind_not_found
       stmt.close
     end
 
+    def test_params
+      stmt = SQLite3::Statement.new(@db, "select  ?1, :foo, ?, $bar, @zed, ?250, @999")
+      assert_equal ["foo", "bar", "zed"], stmt.named_params
```
flavorjones: Related to my previous comment, I think this test needs to be updated and expanded to include numeric names:

```suggestion
      stmt = SQLite3::Statement.new(@db, "select  ?1, :foo, ?, $bar, @zed, ?250, @999, :123, $777")
      assert_equal ["foo", "bar", "zed", "999", "123", "777"], stmt.named_params
```

flavorjones (changes requested):
Please don't worry about the failing tests for windows and ruby 3.2, 3.3, and 3.4 -- they're failing for unrelated reasons.

I added some comments about support for named parameters that happen to be numeric.

This is looking good! I feel like it's really close, just the numeric edge case and I think we can merge it.

PR #656, opened 2025-11-19
Allow calling each, each_hash without block

kch:
Returns an enumerator. Matching similar methods in ruby stdlib.

### DIFF
--- a/test/test_integration_resultset.rb
+++ b/test/test_integration_resultset.rb
@@ -113,15 +113,33 @@ class IntegrationResultSetTestCase < SQLite3::TestCase
   def test_each
     called = 0
     @result.reset(1, 2)
-    @result.each { |row| called += 1 }
+    result = @result.each { |row| called += 1 }
+    result.reset # reset just to confirm we can chain the method after each
+    assert_equal @result, result
     assert_equal 2, called
   end
 
   def test_each_enum
+    @result.reset(1, 2)
+    enum = @result.each
+    assert_instance_of Enumerator, enum
+    assert_equal 2, enum.to_a.length
+  end
+
+  def test_each_hash
     called = 0
     @result.reset(1, 2)
-    @result.each.to_a.each { |row| called += 1 }
+    result = @result.each_hash { |row| called += 1 }
+    result.reset
+    assert_equal @result, result
     assert_equal 2, called
+  end
+
+  def test_each_hash_enum
+    @result.reset(1, 2)
+    enum = @result.each_hash
+    assert_instance_of Enumerator, enum
+    assert_equal 2, enum.to_a.length
   end
 
   def test_enumerable


test/test_integration_resultset.rb
```diff
@@ -139,7 +146,7 @@ def test_close
     assert_predicate stmt, :closed?
     assert_raise(SQLite3::Exception) { result.reset }
     assert_raise(SQLite3::Exception) { result.next }
-    assert_raise(SQLite3::Exception) { result.each }
+    assert_raise(SQLite3::Exception) { result.each.next }
```
flavorjones: Why not also add an assertion for `each_hash.next` while we're here?

test/test_integration_resultset.rb
```diff
@@ -117,6 +117,13 @@ def test_each
     assert_equal 2, called
   end
 
+  def test_each_enum
+    called = 0
+    @result.reset(1, 2)
+    @result.each.to_a.each { |row| called += 1 }
+    assert_equal 2, called
+  end
```
flavorjones: I would like this test to also assert that the `.each` returns an `Enumerator` object. I'd also like a similar test for `.each_hash`, something like:

```ruby
  def test_each_enum
    @result.reset(1, 2)
    enum = @result.each
    assert_instance_of Enumerator, enum
    assert_equal 2, enum.to_a.length
  end

  def test_each_hash_enum
    @result.reset(1, 2)
    enum = @result.each_hash
    assert_instance_of Enumerator, enum
    assert_equal 2, enum.to_a.length
  end
```

flavorjones (commented):
Thanks for opening this PR! This is a nice quality-of-life improvement. I have just a couple of comments about the test coverage (which I have been trying to improve with every PR).

lib/sqlite3/resultset.rb
```diff
@@ -46,6 +46,7 @@ def next
     # Required by the Enumerable mixin. Provides an internal iterator over the
     # rows of the result set.
```
flavorjones: Please also update the docstring!

test/test_integration_resultset.rb
kch: done

lib/sqlite3/resultset.rb
kch: changed. lmk if good.

lib/sqlite3/resultset.rb
```diff
@@ -54,6 +55,7 @@ def each
     # Provides an internal iterator over the rows of the result set where
     # each row is yielded as a hash.
```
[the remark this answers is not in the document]
kch: changed. lmk if good.

flavorjones:
Sorry for the delay here, I've got a large open-source backlog at the moment and plan to get back to this shortly.

PR #669, opened 2025-12-19
Add Database#serialize and Database#deserialize methods

Maaarcocr:
Add support for sqlite3_serialize() and sqlite3_deserialize() which allow serializing an in-memory database to a binary string and restoring it later.

These methods are useful for:
- Creating snapshots of in-memory databases
- Transferring database contents between connections
- Backing up and restoring database state

The methods are only available when SQLite is compiled without SQLITE_OMIT_DESERIALIZE (requires SQLite >= 3.23.0).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,8 +23,8 @@ env:
 jobs:
   ruby_versions:
     outputs:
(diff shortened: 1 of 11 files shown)


Maaarcocr:
closing this in favour of https://github.com/sparklemotion/sqlite3-ruby/pull/685

PR #671, opened 2025-12-23
Fix the inconsistent names in SQLite3::Statement#memused RDoc call-seq

houyuanjie:
It should be `memused`.


flavorjones (approved):
Thank you!

PR #705, opened 2026-05-10
Fix statement_timeout to respect the configured millisecond budget

gauravs:
`statement_timeout=` was aborting on the second progress 
callback regardless of value — `stmt_deadline` was set
to `now` and the configured budget was never added.

Fix: compute `stmt_deadline = now + stmt_timeout` (with 
nsec carry to keep the timespec normalized).

Added `test_statement_timeout_honors_budget_duration` 
which measures elapsed time against 50/100/250 ms budgets. 
Fails on main (elapsed ~0 ms regardless of budget), passes 
here.

## Reproduce

```ruby
# Setup:
#   git clone https://github.com/sparklemotion/sqlite3-ruby
#   cd sqlite3-ruby && bundle install && bundle exec rake compile
#   bundle exec irb
#
# Then paste the rest of this file.
#
# Expected (correct):
#   budget=50ms     elapsed≈50ms
#   budget=500ms    elapsed≈500ms
#   budget=2000ms   elapsed≈2000ms
#
# Actual on main:
#   budget=50ms   elapsed=0.6ms   BUG: aborted before budget
#   budget=500ms   elapsed=0.1ms   BUG: aborted before budget
#   budget=2000ms   elapsed=0.1ms   BUG: aborted before budget

require 'sqlite3'

SLOW_SQL = <<~SQL
  WITH RECURSIVE r(n) AS (
    SELECT 1 UNION ALL SELECT n + 1 FROM r WHERE n < 1000000000
  )
  SELECT count(*) FROM r;
SQL

puts "sqlite3 gem #{SQLite3::VERSION}, sqlite #{SQLite3::SQLITE_VERSION}"
puts

[50, 500, 2000].each do |budget|
  db = SQLite3::Database.new(':memory:')
  db.statement_timeout = budget
  started = Process.clock_gettime(Process::CLOCK_MONOTONIC)
  begin
    db.execute(SLOW_SQL)
    puts "  budget=#{budget}ms   ran to completion (UNEXPECTED)"
  rescue SQLite3::InterruptException
    elapsed_ms = ((Process.clock_gettime(Process::CLOCK_MONOTONIC) - started) * 1000).round(1)
    status = elapsed_ms >= budget * 0.8 ? 'OK' : 'BUG: aborted before budget'
    puts "  budget=#{budget}ms   elapsed=#{elapsed_ms}ms   #{status}"
  end
  db.close
end
```

### DIFF
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -344,6 +344,7 @@ rb_sqlite3_statement_timeout(void *context)
     clock_gettime(CLOCK_MONOTONIC, &currentTime);
 
     if (!timespecisset(&ctx->stmt_deadline)) {
+        // Set stmt_deadline if not already set
         ctx->stmt_deadline = currentTime;
         ctx->stmt_deadline.tv_sec  += ctx->stmt_timeout / 1000;
         ctx->stmt_deadline.tv_nsec += (ctx->stmt_timeout % 1000) * 1000000L;
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -344,8 +344,13 @@ rb_sqlite3_statement_timeout(void *context)
     clock_gettime(CLOCK_MONOTONIC, &currentTime);
 
     if (!timespecisset(&ctx->stmt_deadline)) {
-        // Set stmt_deadline if not already set
         ctx->stmt_deadline = currentTime;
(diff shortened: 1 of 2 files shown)


gauravs:
Found a possible issue with how it works with AR. Closing this for now and will redo once I have a better solution.

flavorjones:
@gauravs Feel free to open an issue describing the problem, if you don't have time to work on a fix.

PR #710, opened 2026-06-07
Fix use-after-free issue with custom functions

flavorjones:
when the same function name is used with multiple arities.

ref: https://github.com/sparklemotion/sqlite3-ruby/security/advisories/GHSA-28hh-pr2h-2w89

### DIFF
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -528,7 +528,7 @@ define_function_with_flags(VALUE self, VALUE name, VALUE flags)
 
     CHECK(ctx->db, status);
 
-    rb_hash_aset(rb_iv_get(self, "@functions"), name, block);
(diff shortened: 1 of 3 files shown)


maikelvdh:
@flavorjones is it possible by chance get this vulnerability also fixed in v1.7.x version range under a new 1.7.4 release? thanks in advance.

flavorjones:
@maikelvdh Thanks for asking, but I'm not willing to support the 1.7 branch, which has not been touched in over two years. And for what it's worth, I do not consider this to be a vulnerability, since it requires a developer to write specific code (and cannot be triggered by untrusted input). If this fix is important to you, though, you should be able to apply the patch to te 1-7-stable branch with onlysmall changes and run your application on that fork.

PR #719, opened 2026-07-15
Close connection immediately after failed open

katafrakt:
Per [documentation](https://www.sqlite.org/c3ref/open.html) `sqlite3_close_v2` should be called to free up resources even after `sqlite3_open` (and family) fails:

> Whether or not an error occurs when it is opened, resources associated with the [database connection](https://www.sqlite.org/c3ref/sqlite3.html) handle should be released by passing it to [sqlite3_close()](https://www.sqlite.org/c3ref/close.html) when it is no longer required.

This is a bit verbose, because we need to get the message before closing and pass to `CHECK_MSG`. Alternative would be to keep using `CHECK`, but then a generic message would be printed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # sqlite3-ruby Changelog
 
+## next / unreleased
+
+### Improved
+
+- When `Database.new` fails to open the database file, the underlying sqlite3 connection handle is now closed immediately instead of waiting for the garbage collector to clean it up. #719 @katafrakt
+
+
 ## 2.9.5 / 2026-06-07
 
 ### Dependencies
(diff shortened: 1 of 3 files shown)


flavorjones:
@katafrakt Thanks for the PR. Reading the docs, this does sound like the right thing to do.

Can I ask how you ran into this? Is there an easy way to reproduce the underlying failure conditions? I'd like to try to write a test and run it under valgrind/ASan to make sure memory leaks are all quashed.

flavorjones:
Ah - ignore my question, I see it's as easy as opening a nonexistent file. I'll try to put a test around this, but the change itself seems fine. Thanks!

Worth noting: in the current codebase the memory is _eventually_ cleaned up when the object is GCed, but this change would make that cleanup happen immediately which is definitely better.

katafrakt:
The story here is that I did not encounter this problem per se. I was making a similar fix for Elixir driver and checked how other drivers (Ruby, Python) handle this. I found out that Ruby has a similar problem as Elixir and decided to make a PR here as well.

PR #737, opened 2026-08-25
Fix statement_timeout interrupting most queries

palkan:
When statement_timeout is set, the queries triggering the progress handler more than once are interrupted regardless of the timeout value: the current implementation just doesn't use the configured timeout and set the deadline to currentTime (see the diff).

Fast queries still complete (because they trigger the progress handler only once), so it may seem to work but in reality it doesn't. (I caught this bug while writing a test confirming the timeout works in my app; haven't seen in prod env though.)

(A similar bug/fix was reported/proposed earlier in #705).

### DIFF
--- a/ext/sqlite3/database.c
+++ b/ext/sqlite3/database.c
@@ -417,8 +417,10 @@ rb_sqlite3_statement_timeout(void *context)
     clock_gettime(CLOCK_MONOTONIC, &currentTime);
 
     if (!timespecisset(&ctx->stmt_deadline)) {
-        // Set stmt_deadline if not already set
-        ctx->stmt_deadline = currentTime;
+        struct timespec timeout;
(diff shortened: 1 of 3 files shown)


flavorjones:
Thanks for the pull request. I'll review as soon as I get a chance.

Can you please consider whether we need a specific new test for this behavior? I haven't formed an opinion so your feedback would be welcome, and "no" might be a fine answer.

palkan:
@flavorjones I think, we missed one check (added): setting timeout to 0 should allow the original interrupted query to complete. Added timings assertions, too (just in case our timeout calculation is not what it says).

Also, tweaked the numbers a bit to reduce flakiness (10ms timeout could be too small to trigger the second progress handler invocation and cause the interruption). (That's why a couple of jobs failed the last time.)

